PR #6, opened 2015-04-03
Updated to 3C fonts

neolinkpsu:
- Added in new font icons
- Updated SingleNickel icons to use the new fonts
- Updated some of the SingleNickel buttons


spullen:
:+1:

PR #8, opened 2015-04-07
TC-1401 Incorrect adjacencies heatmap

spullen:
https://thirdchannel.atlassian.net/browse/TC-1401


js/apps/thirdchannel/init.js
```diff
@@ -29,6 +29,7 @@ require.config({
         'helpers' : 'thirdchannel/utils/helpers',
         'pikaday' : '../libs/bower_components/pikaday/pikaday',
         'chartist' : '../libs/bower_components/chartist/dist/chartist',
+        'd3': '../libs/bower_components/d3/d3.min',
```
neolinkpsu: Change this to the development d3 (without min).  Our requireJs will minimize it when it compiles everything together.  Also makes it easier for us to debug issues happening inside d3

js/apps/thirdchannel/views/reports/widgets/heatmap.js
```diff
@@ -0,0 +1,166 @@
...
+            this.model.config.legend = legend;
+
+            this.loadingView = new LoadingView();
+
+            if(window.pdf === undefined) {
+                this.listenTo(context, 'report post render', _.debounce(function() {
+                    self.loadingView.remove();
+                    self.renderChart();
+                }, 500));
+            } else {
```
neolinkpsu: Why is this not also true for pdf?

js/apps/thirdchannel/views/reports/widgets/heatmap.js
```diff
@@ -0,0 +1,166 @@
+define(function(require) {
+    var Backbone = require('backbone'),
+        Handlebars = require('handlebars'),
+        HandlebarsTemplates = require('handlebarsTemplates'),
+        d3 = require('d3'),
+        context = require('context'),
+        LoadingView = require('thirdchannel/views/utils/loading');
+
+    var rowLabelMargin = 125;
+    var colLabelMargin = 130;
```
neolinkpsu: what are these doing?

js/apps/thirdchannel/views/reports/widgets/heatmap.js
spullen: I added a loadingView, which I then remove once the nav and filter have transitioned. However, the pdf does not like the loadingview, or really I should say the pdf view doesn't have the same problems with the nav and filter, therefore it can just render. If the loadingview is there, that's what ends up getting displayed instead of the chart. If I use the debounce logic, sometimes it doesn't render the chart... The struggle was real.

js/apps/thirdchannel/views/reports/widgets/heatmap.js
spullen: These are used to make room for the labels. Wasn't able to find a way to calculate them prior to appending the labels into the svg. 

I'll take another stab at this (obviously this won't scale), and I'm sure there's a way.

PR #9, opened 2015-04-09
Added in debouncing for event calls to render the page.  It should now o...

neolinkpsu:
...nly be calling rendering on most elements once.


spullen:
:+1:

PR #11, opened 2015-04-16
Fixed the way the legends were being created since this was causing issu...

neolinkpsu:
...e with lines of more than 2 datasets.


js/apps/thirdchannel/views/reports/widgets/line_chart.js
```diff
@@ -88,26 +88,20 @@ define(function(require) {
         },
         setupColors: function() {
             var self = this,
-                total_entries = this.model.results.datasets.length,
-                setLegendColors = true;
+                total_entries = this.model.results.datasets.length;
```
spullen: Looks like we can remove this since it's no longer being used

spullen:
:+1:

PR #12, opened 2015-04-16
TC-1425 - Heat map view breakdowns

spullen:
https://thirdchannel.atlassian.net/browse/TC-1425

See companion PR on 3C side

### DIFF
--- a/js/apps/thirdchannel/views/reports/widgets/heatmap.js
+++ b/js/apps/thirdchannel/views/reports/widgets/heatmap.js
@@ -9,12 +9,14 @@ define(function(require) {
     return Backbone.View.extend({
         template: HandlebarsTemplates['thirdchannel/reports/widgets/heatmap'],
         initialize: function (options) {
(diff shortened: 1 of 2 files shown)


js/apps/thirdchannel/views/reports/widgets/heatmap.js
```diff
@@ -126,6 +129,25 @@ define(function(require) {
...
+                            viewBreakDownLink += '&' + param + '=' + val;
+                        });
+                    }
+
+                    //context.router.navigate(viewBreakDownLink, {trigger: true});
+                    // ugh...
+                    window.location = viewBreakDownLink;
+                });
+            }
+
```
neolinkpsu: the query string passed in should already be formatted for you.  Why are you having to format it?

spullen: So the reason I did this was because the customer_account was included in the link, which needs to be replaced. However, I tried it without any modifications and it seemed to work, I think the tacked on version takes precedence.

PR #14, opened 2015-04-22
TC-1084

spullen:
https://thirdchannel.atlassian.net/browse/TC-1084


neolinkpsu:
Just need to check that nothing is broken on beta on mobile and if that is working, I :+1:

spullen:
Pushed to staging, everything seems to be in working order.

PR #17, opened 2015-05-18
Optimized the heatmap class.

neolinkpsu:
- Removed unnecessary function calls
 - Fixed issue where the axis' labels would change if resized.
 - Fixed an issue with the legend being offcenter when in mobile.
 - Reduced code to just a few standard calls.


spullen:
:+1:

PR #18, opened 2015-05-20
TC-1500

spullen:
Image grouping client side stuffz


js/apps/thirdchannel/views/s3uploader/checkin_file.js
```diff
@@ -0,0 +1,118 @@
...
+                // ...
+                m.beforeImages = self.model.beforeImages;
+
+                new ImageView({model: m}).setElement(this);
+            });
+        },
+        render: function() {
+            return this;
+        },
+        fileChanged: function(e) {
```
spember: Have you thought about potentially extracting the code around File Handling into it's own View? If we ever need to upload additional files, it may be useful.

no need to do this now, of course

spember:
I see no problems! :+1: :shipit:

js/apps/thirdchannel/views/s3uploader/checkin_file.js
spullen: Yup, that's a good suggestion. I think the original intent was for the file view to be generic, but with this new thing I was trying to fit it I had to bastardize it a bit. I'll give a little thought in how I might be able to bring it back, and if I can get it in quickly then I implement it.

PR #20, opened 2015-06-05
TC-1523

spullen:
* Removes the re-calculate mileage feature since the calculation has been simplified.


spember:
lgtm :rocket:

spullen:
Not sure what this was, but probably irrelevant at this point.

PR #21, opened 2015-06-10
TC-1552 Adding estimated travel expenses for a schedule

spullen:
PMs should be able to see an estimate of travel expenses for current and upcoming schedules.


js/apps/pennyPacker/views/entries/actions.js
```diff
@@ -34,42 +24,7 @@ define(function(require) {
                 e.stopPropagation();
                 e.preventDefault();
                 window.location = context.content.links.paypal + window.location.search + "&format=csv";
-            },
-
-            findMissingMileage: function () {
-                var $span = this.$el.find('.missing-mileage span');
```
matthewlittle: You can shortcut that find by:
```
 this.$('.missing-mileage span');
```

matthewlittle:
look good.

PR #26, opened 2015-06-15
Tc 1646

neolinkpsu:
Added in additional properties for the checkin details to be displayed.
 - Tier information
 - Survey name
 - Link to checkin repor
 - Link to store profile


spullen:
:+1:

PR #27, opened 2015-06-19
TC-1663 ftue cleanup

spullen:
* Changes the path that triggers the shipping form validations during the FTUE process

See thirdchannel and supplication PRs


neolinkpsu:
:+1:

PR #28, opened 2015-06-29
TC-1681 - Adding interface to manage missing travel entries for visits

spullen:
* Adds a form to select a date range to find visits that don't have travel
* Displays a list of visits where the user can click to create the travel entry.

Also see PennyPacker PR https://github.com/RSV2/PennyPacker/pull/7


js/apps/pennyPacker/init.js
```diff
@@ -18,8 +18,7 @@ require.config({
         'helpers'       : 'thirdchannel/utils/helpers',
         'pikaday'       : '../libs/bower_components/pikaday/pikaday',
         'context'       : 'shared/utils/context',
-        
-
+        'noty'          : '../libs/bower_components/noty/js/noty/packaged/jquery.noty.packaged'
```
spember: what does noty do for us?

spullen: Notifications for flash messages. Just to provide feedback to the user (something we need to do more of).

spember: nice

spember:
lgtm

PR #29, opened 2015-07-09
TC-1702 - Adds a view to see all the scheduling cycles ordered by start date descending

spullen:


js/apps/procrastination/models/admin/scheduling_cycle.js
```diff
@@ -0,0 +1,7 @@
+define(function(require) {
+    var Backbone = require('backbone'),
+        context = require('context');
+
+    return Backbone.Model.extend({
```
spember: What is the point to this model? Seems unnecessary

spember:
seems fine to me, although I'm curious what the need for a specific SchedulingCycle model is? Could you get by with just a 'Backbone.Model'... or, rather, if no model is set to a BB collection, it simply wraps all objects in a Backbone.Model

js/apps/procrastination/models/admin/scheduling_cycle.js
spullen: True, at one point it did have some other functionality, but it was abandoned and I was being lazy and didn't remove it.

PR #30, opened 2015-07-10
TC-1147 - Filterable filters

spullen:
Adds the ability to filter down a filter list. Works around how the filter works (hide/show of active/inactive items).


neolinkpsu:
Code looks alright.  Lets get this up on Beta so I can play around with it to see if there are any issues.

PR #34, opened 2015-08-20
TC-1869 - Updating references to customer store using uuid instead of id

spullen:
Updating references to customer store using the UUID instead of the ID. There weren't too many changes here since we do a good job of either pulling out the id from the param arguments by location (location didn't change just the value did, and the value is set via the rails app code) or using links in the json payloads.

Also see the thirdchannel app PR.


spullen:
Closing this PR, since there were more changes relating to this on other PRs rolling all of it up into one.

PR #35, opened 2015-08-24
Added the ability to show all of the choices of a question through a …

neolinkpsu:
…modal. (TC-1858)  This is Part 1 of 2 for this ticket. The other is in the ThirdChannel app.

### DIFF
--- a/js/apps/thirdchannel/models/reports/checkins/choices.js
+++ b/js/apps/thirdchannel/models/reports/checkins/choices.js
@@ -1,0 +1,12 @@
+define(function(require) {
+    var Backbone = require('backbone');
+
+    return Backbone.Model.extend({
+        initialize: function (options) {
+            this.options = options;
+        },
+        url: function () {
+            return '/programs/' + this.options.programId + "/reports/checkin/" + this.options.checkinId + "/question/" + this.options.questionId + "?format=json";
+        }
+    });
+});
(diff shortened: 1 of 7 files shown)


js/apps/thirdchannel/models/reports/checkins/choices.js
```diff
@@ -0,0 +1,12 @@
+define(function(require) {
+    var Backbone = require('backbone');
+
+    return Backbone.Model.extend({
+        initialize: function (options) {
+            this.options = options;
+        },
+        url: function () {
+            return '/programs/' + this.options.programId + "/reports/checkin/" + this.options.checkinId + "/question/" + this.options.questionId + "?format=json";
```
spullen: If you passed in the attributes to the model like `var model = new choicesModel({programId: someProgramId, ...})` you should be able to get that value like `this.get('programId')` and you can get rid of the constructor if you did it this way.

neolinkpsu: Once you do a fetch those attributes go away.  It's best to try and keep these things as members of the model and out of the attributes area.

spullen: Ah, yeah I see the problem now. The model being fetched probably should have those attributes being returned as json so we can use backbone more purely, but for now this works.

spullen: Actually not sure if that's true (or maybe it's just the version of backbone we're using), check this out: http://jsbin.com/kajubu/1/edit?html,js,output (may need to open up the real console to see it). I added some random attributes that are not in the returned attribute set and they were still there after the data was fetched.

PR #37, opened 2015-09-08
Optional images

neolinkpsu:
This contains source code for about 3 tasks.
- First is the summary question on the survey builder.  Since this question type will be coming sometime soon, I don't think we need to worry about extracting it out, since we don't have any surveys we're building in the near future.
- Second is the fixes for a collapse and expand buttons on the reports.  Quick fix.
- Third is the optional images.  I've expanded the valid method to look for the existence of a required before and after image.  If they exist than they are checked for the usual things, otherwise we allow the user to enter an image if they want in either the before or after areas.


spullen:
Looks good.

PR #38, opened 2015-09-10
tc-1939

matthewlittle:
Updated the filterableComponent to use a data-filter field rather than the data-value. This new field is equal to the value viewed by users making the filter work on what the user sees, not the value of the form field.


spullen:
looks good to me.

PR #39, opened 2015-09-22
TC-1961 - Adding the ability to export survey answers

spullen:
https://thirdchannel.atlassian.net/browse/TC-1961

Adds the ability to download answers for a given survey.

Made the existing export modal a little more flexible in order to handle this new export and the existing report one.

Also see ThirdChannel PR.


neolinkpsu:
:-1:   I believe you've broken csv export on the view breakdown pages

spullen:
Ok, will check out

spullen:
Ok, fixed now. Was able to reuse/cleanup the report export model and views for both since I inverted how the model gets assigned to the view.

PR #41, opened 2015-10-02
TC-1993 - displaying the agent's program profile store list

spullen:
Front end for the profile store list (using the newer pagination scheme).

See PRs for supplication and thirdchannel.

### DIFF
--- a/js/apps/thirdchannel/routers/router.js
+++ b/js/apps/thirdchannel/routers/router.js
@@ -29,13 +29,15 @@ define(function(require){
         NotificationBadge = require('thirdchannel/views/notifications/notification_badge'),
         StoreProfileSchedule = require('thirdchannel/views/store_profile/schedule'),
         Labs = require('thirdchannel/views/labs/main'),
-        AnswersExportView = require('thirdchannel/views/exports/answers/main');
+        AnswersExportView = require('thirdchannel/views/exports/answers/main'),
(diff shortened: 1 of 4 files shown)


js/apps/thirdchannel/routers/router.js
```diff
@@ -149,6 +151,10 @@ define(function(require){
             ActivitiesMain.init(url, null, false);
         },
 
+        programProfileStores: function(program_id, user_id) {
+            new ProfileStoreListView().bootstrapCollection(window.bootstrap);
```
spullen: So Filter.init() didn't fix that issue that we were talking about on our way over to grab coffee.

js/apps/thirdchannel/views/profiles/stores/list.js
```diff
@@ -0,0 +1,12 @@
+define(function(require) {
+    var context = require('context'),
+        PageableListView = require('thirdchannel/views/shared/pageable_list'),
+        StoreListView = PageableListView.extend({
+            el: '#stores',
+            template: 'thirdchannel/profiles/stores/rows',
+            pageChange: function (page) {
+                context.trigger('filter:query', "page=" + page);
```
spullen: workaround for triggering the page change, this eventually changes the queryString on the collection object, which is then reset from server.

neolinkpsu:
:+1:

PR #42, opened 2015-10-05
tc-1965: changes to support new TOMS widgets

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-1965

Added the ability to make a metric icon show "No Data" if it is in a given state

Added a config option for limiting the y tick


spullen:
looks good.

PR #44, opened 2015-10-30
TC-2000 - Updates to the Report Info Lists

spullen:
https://thirdchannel.atlassian.net/browse/TC-2000

* Updated how the data was being structured so it's a little easier to use in the view
* Added the task to be displayed above the answer (helpful for programs like luxottica where we can have multiple answers but it's impossible to know which survey (brand) they were in response to).


spember:
:+1: :shipit:

PR #45, opened 2015-10-30
tc-2068: show/hide explain field based on trigger

eliottwiener-thirdchannel:
Hide or show the explain field based on whether trigger options are selected onchange.

https://thirdchannel.atlassian.net/browse/TC-2068

### DIFF
--- a/js/apps/thirdchannel/views/checkins/show/survey.js
+++ b/js/apps/thirdchannel/views/checkins/show/survey.js
@@ -95,9 +95,9 @@ define(function(require) {
         },
         displayQuestionExtraIfTriggered: function(e) {
           if(this.$(e.currentTarget).find(":selected[data-trigger=true]").length > 0){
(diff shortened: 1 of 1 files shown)


js/apps/thirdchannel/views/checkins/show/survey.js
```diff
@@ -92,6 +93,13 @@ define(function(require) {
         hideElement: function(e) {
             this.$(e.currentTarget.getAttribute('data-hide-element')).hide('fast', "linear").val('').trigger('change');
         },
+        displayQuestionExtraIfTriggered: function(e) {
+          if(this.$(e.currentTarget).find(":selected[data-trigger=true]").length > 0){
+            this.$(e.currentTarget.getAttribute('data-show-element')).show('fast', "linear");
```
spullen: I still think we should just change this data attribute to something else, especially since it's just being duplicated twice on the element with the same information. I'd call it `data-trigger-element` and use it for both the show/hide parts of the if statement.

PR #46, opened 2015-11-03
TC-2131 - updates views 

spullen:
https://thirdchannel.atlassian.net/browse/TC-2131

* Updates views for oddjob and procrastination with information from store service (basically some variable changes).

Also see PRs for ThirdChannel, Oddjob, Store Service, and Procrastination


spember:
lgtm :shipit:

spember:
Did you verify that you can add stores to a job in oddjob? If so, this, and the oddjob PRs both get a :+1:

spullen:
Yes, I did test the add/removal of stores from a job.

Probably will have everyone get involved with smoke testing on staging prior to releasing since it touches a few critical path areas.

PR #48, opened 2015-12-01
tc-2347: Add support for View Breakdown links to range chart

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-2347


spullen:
Looks good to me.

PR #49, opened 2015-12-02
Fixed the issue with the ROI export and get it inline with the standa…

neolinkpsu:
…rd csv export.  Part 2 of 2.  First part is in ThirdChannel

Just added a view to the ROI page which will perform the async export


js/apps/thirdchannel/views/exports/base.js
```diff
@@ -0,0 +1,43 @@
+define(function(require) {
```
spember: Can this be reused in other places we do async export generation?

spember:
lgtm :ship:

PR #50, opened 2015-12-02
TC-2037 - Adds the ability to remove a program store from the interface

spullen:
https://thirdchannel.atlassian.net/browse/TC-2037

This is the case where they did not mean to upload a store, or they uploaded incorrect data. In this instance they had given us a store list a while back where Golf Galaxy's (which were already in the system) were mislabeled as Dick's Sporting Goods.

* Adds interface to remove store, adds confirmation prompting the user to enter the store name in order to move forward 

Also see PRs for thirdchannel, supplication, and store service.


js/apps/stores/views/program_stores/delete_modal.js
```diff
@@ -0,0 +1,75 @@
...
+                        speed: 500
+                    },
+                    timeout: 2500
+                });
+
+                this.triggerCancel();
+            }.bind(this)).fail(function(response) {
+                var error = JSON.parse(response.responseText);
+
+                noty({
```
spember: nice. noty.

spullen: yeah, really nice library, we need to use it everywhere!

templates/handlebars/stores/program_stores/delete.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="bbm-button cancel-button pull-right"><i class="ic ic_x"></i></div>
+<div class="delete-confirmation-modal">
+    <h2>Remove {{name}}?</h2>
+    <p class="danger">
+        Danger: Removing a store from a program is a dangerous operation. Stores can be used in multiple locations in the system. Please understand this can potentially cause issues.
+    </p>
+    <div>
+        <label for="name-confirmation">Please enter the name of the store to confirm the delete:</label>
```
spember: I like this: big warning and then make them type instead of clicking

spember:
looks good :shipit:

PR #53, opened 2015-12-08
TC-2045 - export of sales store data

spullen:
https://thirdchannel.atlassian.net/browse/TC-2045

Just uses the same pattern that we've been using for other long running exports

Also see PRs for thirdchannel and amalgamation


neolinkpsu:
I left a comment about how I made a bunch of generic models, views, controllers for the ROI export.  Take a look at how I did that and see if you can reuse it.

neolinkpsu:
:+1:

PR #55, opened 2015-12-08
Adds Async Loading of report widgets to the Labs page

spember:
Adds a new view (async_report.js), which, when given a meta-data of a report (available via the report_meta_data method in reportService) and a widget fetch url (there's a new widgets controller for this purpose, although labs uses a slightly different endpoint, but the same underlying service call), will render the sections, subsections, and then asynchronously load the widgets and insert them in the proper subsection at the proper index.

This should be reusable for the general reports.

I'm not particularly happy with the _renderWidget method in async_report, specifically the bit that determines where to position the widget (index). Necessary as widgets can potentially be fetched out of order. Suggestions welcome.


js/apps/thirdchannel/views/labs/sc/side.js
```diff
@@ -1,105 +1,112 @@
...
+         *  
+         * @type View
+         */
+        SalesCompareSideView = {
+            initialize: function (opts) {
+                if (opts.groupSelect === undefined) {
+                    throw "No 'groupSelect' parameter set in constructor for the SalesCompareSideView";
+                }
+                this.$groupSelect = opts.groupSelect;
+                this.loadingHTML = this.$el.find(".loader")[0];
```
spullen: You can use the shorthand for querying for dom elements within a backbone view (if you want)

```javascript
this.loadingHTML = this.$('.loader')[0];
```

js/apps/thirdchannel/views/labs/sc/side.js
```diff
@@ -1,105 +1,112 @@
...
+
+
+            applyFilter: function (qs) {
+                this.$el.html(this.loadingHTML);
+                this.currentQS = qs;
+                qs = qs + "&group=" + encodeURIComponent(this.$groupSelect.val());
+                this.model.setQueryString(qs);
+                this.model.fetch()
+                 .fail(function () {
+
```
spullen: Should we display a message on failure? maybe introduce noty to the thirdchannel app?

js/apps/thirdchannel/views/reports/async_report.js
```diff
@@ -0,0 +1,125 @@
...
+                var self = this;
+                _.each(this.report.get('sections'), function(section) {
+                    self._layoutSection(section);
+                });
+            },
+
+            loadWidgets: function (filter) {
+                var self = this,
+                    promise = $().promise();
+                _.each(self._extractWidgetMetaInfo(this.report), function(widget) {
```
spullen: I wonder if this can be cleaned up by using `$.when`

For instance
```javascript
var promises = _.map(self._extractWidgetMetaInfo(this.report), function(widget) {
  return $.getJSON(...).then(function(data) { /* render widget */ });
});
$.when(promises).done(function() {
  // remove spinner
  // trigger complete
});
```

js/apps/thirdchannel/views/labs/sc/side.js
```diff
@@ -1,105 +1,112 @@
...
+                this.$el.html("");
+                this.reportLoader.layout();
+                this._renderMeta();
+                this._renderSales();
+                this.reportLoader.loadWidgets(this.model.queryString);
+                
+                return this;
+            },
+
+            updateLinks: function (e) {
```
spullen: when does this get called? when something like a view breakdown is clicked?

js/apps/thirdchannel/views/reports/async_report.js
```diff
@@ -0,0 +1,125 @@
...
+
+            _layoutSection: function(section) {
+                section.title = section.name;
+                var $section = $(Templates['thirdchannel/reports/index/section'](section));
+
+                _.each(section.subsections, function(subsection) {
+                    subsection.title = subsection.name;
+                    var $subsectionsContainer = $section.find('.subsections');
+                    $subsectionsContainer.append(Templates['thirdchannel/reports/index/subsection'](subsection));
+                    $subsectionsContainer.append("<i class='fa fa-spin fa-spinner fa-2x widget-spinner'></i>");
```
spullen: Do we have a spinner template? I believe Andrew just changed this and this spinner might not be the same one he changed it to.

js/apps/thirdchannel/views/reports/async_report.js
spember: I'll try it out, was unsure if when accepted an array

js/apps/thirdchannel/views/labs/sc/side.js
spember: updateLinks? Yeah, because we use relative urls for the breakdown links, it has to rewrite the url on the fly. ugh.

js/apps/thirdchannel/views/labs/sc/side.js
spember: meh

js/apps/thirdchannel/views/labs/sc/side.js
```diff
@@ -1,105 +1,112 @@
...
+                qs = qs + "&group=" + encodeURIComponent(this.$groupSelect.val());
+                this.model.setQueryString(qs);
+                this.model.fetch()
+                 .fail(function () {
+                        
+                 });
+            },
+
+            render: function () {
+                this.$el.html("");
```
spullen: This works, but why not use `this.$el.empty()`

js/apps/thirdchannel/views/reports/async_report.js
spullen: If passing an array directly doesn't work, how about `$.when.apply($, promises)`

spember: :rage:  d'oh. of course

js/apps/thirdchannel/views/reports/async_report.js
spember: You're right, and I'll look into it. Changing this was Brendan's feedback, too.

js/apps/thirdchannel/views/labs/sc/side.js
spember: because I forgot

spullen:
:+1:

PR #56, opened 2015-12-18
tc-2589: async widget loading on reports

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-2589

related to: https://github.com/RSV2/ThirdChannel/pull/188


spullen:
Looks good to me :+1:

PR #57, opened 2015-12-23
tc-2648: load widgets in order

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-2648

The code that inserted widgets in their idx-dependent order was not always working right. I changed the way the async_report view works in order to make it consistent. Now, when the report_meta is render as the report skeleton, hidden placeholders are created for each widget. When each widget is rendered, the placeholder is swapped out for the actual widget view.


spullen:
looks good to me :+1:

PR #59, opened 2016-01-06
tc-2706: load up a new report_meta when the filter changes

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-2706

see also: https://github.com/RSV2/ThirdChannel/pull/192

When the filter was changing, the report was staying the same. Now when the filter changes, we should get a new report_meta before starting the async widget loading process.


spullen:
Looks fine to me.

PR #61, opened 2016-01-08
tc-2712: separate synchronous view for pdf reports

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-2712

https://github.com/RSV2/ThirdChannel/pull/193


neolinkpsu:
:+1:

PR #63, opened 2016-01-18
tc-2774: if report param is not present, just get the default report

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-2774

I forgot to include the case where the "report" param was left blank when loading a different report


spullen:
Looks fine to me

PR #64, opened 2016-01-20
tc-2610: new, more hierarchical checkins page

eliottwiener-thirdchannel:
Instead of having tasks directly under stores, we will now list jobs under stores, and then tasks under jobs. We show the date a job is scheduled for, if applicable (agent checkin).

see also:
https://thirdchannel.atlassian.net/browse/TC-2586
https://github.com/RSV2/Procrastination/pull/9
https://github.com/RSV2/ThirdChannel/pull/246


js/apps/thirdchannel/views/checkins/job.js
```diff
@@ -0,0 +1,61 @@
...
+        closeBtn: "Hide Tasks <i class='ic ic_up'></i>",
+        events: {
+            "click .open-tasks": "openTasks",
+            "click .close-tasks": "closeTasks"
+        },
+        initialize: function(options){
+            if(options.model.date !== ""){
+                // parse date (iso8601 format) as a date in utc
+                // and format it as a localized date in utc
+                this.model.local_date = moment.utc(options.model.date).format("l");
```
eliottwiener-thirdchannel: Should dates be represented in the UI with local tz or as UTC dates?

spullen: Displaying it in the user's local time would be ideal.

spullen: As we talked about earlier on slack, probably makes sense to use the same thing that the scheduling service does to display dates.

Also, this seems like it would be better suited as a handlebar helper method to use within the template instead of here.

js/apps/thirdchannel/views/checkins/job_list.js
```diff
@@ -0,0 +1,23 @@
...
+    var Backbone = require('backbone'),
+        context = require('context'),
+        JobView = require('thirdchannel/views/checkins/job'),
+        HandlebarsTemplates = require('handlebarsTemplates');
+
+    return Backbone.View.extend({
+        className: 'job-list',
+        loadingHTML: HandlebarsTemplates['thirdchannel/loading_icon'],
+        render: function() {
+            var dates = _.keys(this.model.jobs_by_date).sort();
```
spullen: I feel like we're doing a lot of work to massage the data on the frontend. Could this be cleaned up on the backend?
Maybe structure it like (this is hypothetical and is not one to one matching the fields we have)
```
{
  [
    {
        program_store_uuid: <uuid>,
        store_name: <name>,
        store_address: '123 Fake St',
        jobs: [
            {
               job_uuid: <JOB_UUID>,
               date: '2016-01-20',
               tasks: [
                  {
                      task_id: <TASK_ID>,
                      survey_uuid: <SURVEY_UUID>
                      task_name: 'Do the thing'
                  }, { task_id: <TASK_N>, ...}
               ]
            },
           { ... }
        ]
    },
   { store_n... }
  ]
}
```
Something like this would clean up the front end a lot cleaner

eliottwiener-thirdchannel: I didn't structure it that way because I was trying to avoid having copies of data, but I ended up having copies of tasks anyway. I would be open to cleaning it up and using a structure like that to make the code easier to follow.

spullen: I think this should be done on the backend (even if there's some duplication (which won't be a lot anyways)). The front end should be really straightforward.

spullen: You can hold off on doing anything about this. I need to think about it a little more.

eliottwiener-thirdchannel:
This is ready for review again

css/scss/apps/thirdchannel/_checkin.scss
```diff
@@ -1,30 +1,34 @@
...
+	font-style: italic;
+	font-size: 0.8rem;
 }
 
-#taskContainer {
-  .task-item, .complete-checkin-tasks {
-    padding-bottom: 1em;
-    padding-top: 1em;
-  }
+.error {
```
spullen: This seems pretty generic, there's probably a better place to put this (modules/atoms perhaps).

Also might want to check for an existing error class, we may already have one.

js/apps/thirdchannel/routers/router.js
```diff
@@ -135,6 +137,10 @@ define(function(require){
             CheckinsView.init();
         },
 
+        in_progress: function (program_id, checkin_id){
```
spullen: Minor thing, but can we use camel case (function name and variables) to keep it in line with the js code style.

js/apps/thirdchannel/views/checkins/expandable.js
```diff
@@ -0,0 +1,56 @@
...
+        className: 'expandable pure-g',
+        events: {
+            "click .open": "opensubsection",
+            "click .close": "closesubsection"
+        },
+        initialize: function(){
+            this.listenTo(this, 'subsection:open', this.opensubsection);
+        },
+        render: function() {
+            this.$el.html(HandleBarsTemplates['thirdchannel/checkins/expandable']({
```
spullen: Why not set the template variable? That'll clean this up a little.

js/apps/thirdchannel/views/checkins/expandable.js
```diff
@@ -0,0 +1,56 @@
...
+                toggle: this.openText + "&nbsp;<i class='ic ic_down'></i>",
+                hide_toggle: this.model.hide_toggle,
+            }));
+            this.main = this.$el.find(".main");
+            this.main.html(HandleBarsTemplates[this.template](this.model));
```
spullen: Where is `this.template` being set? And it seems like it's not using it correctly in the backbone sense.

eliottwiener-thirdchannel: There are a few views that extend expandable.js that will set `this.template`.

spullen: ok, seeing that now. Could have a different variable name to reduce the confusion.

js/apps/thirdchannel/routers/router.js
eliottwiener-thirdchannel: The rest of the file uses snake_case already, so it's consistent this way.

spullen: I see about five examples of snake case (two of which are functions in_progress and checkin_edit), definitely not the majority.

js/apps/thirdchannel/routers/router.js
```diff
@@ -133,10 +135,14 @@ define(function(require){
             new AdminView();
         },
 
-        checkin_list: function (program_id, user_id){
+        checkinList: function (){
             CheckinsView.init();
         },
 
+        inProgress: function (){
```
spullen: Thanks!

js/apps/thirdchannel/views/checkins/date.js
```diff
@@ -0,0 +1,23 @@
...
+        HandlebarsHelpersExt = require('handlebarsHelpersExt'),
+        Job = require('thirdchannel/views/checkins/job');
+
+    return Backbone.View.extend({
+        className: 'visit section',
+        template: HandlebarsTemplates['thirdchannel/checkins/date'],
+        render: function() {
+            this.$el.html(this.template(this.model));
+            this.jobs = this.$el.find(".jobs");
+            this.jobs.append("<div>Jobs:</div>");
```
spullen: Why isn't this just part of the template?

js/apps/thirdchannel/views/checkins/date.js
```diff
@@ -0,0 +1,23 @@
...
+    var HandlebarsTemplates = require('handlebarsTemplates'),
+        HandlebarsHelpersExt = require('handlebarsHelpersExt'),
+        Job = require('thirdchannel/views/checkins/job');
+
+    return Backbone.View.extend({
+        className: 'visit section',
+        template: HandlebarsTemplates['thirdchannel/checkins/date'],
+        render: function() {
+            this.$el.html(this.template(this.model));
+            this.jobs = this.$el.find(".jobs");
```
spullen: Not necessary, but it's good to annotate variables that reference dom elements using the `$`, in this example `this.jobs = ...` would be `this.$jobs = ...`.

Also, just so you're aware, not necessary to correct, `this.$('.jobs')` would be shorthand for `this.$el.find('.jobs')`.

js/apps/thirdchannel/views/checkins/date.js
```diff
@@ -0,0 +1,23 @@
...
+            this.jobs.append("<div>Jobs:</div>");
+            _.each(this.model.job_uuids, function(job_uuid){
+                var jobView = new Job({model: {
+                    job: this.model.job_details_by_uuid[job_uuid],
+                    hide_toggle: true,
```
spullen: Seems like this is just internal state to the expandable view. Is there any reason why hide_toggle can't be instantiated in that class? If it can't, it seems more fitting to be an option versus a model, `var jobView = new Job({model: this.model.job_details_by_uuid[job_uuid], hide_toggle: true})`, then inside expandable:
```javascript
...
initialize: function(options) {
  this.hide_toggle = options.hide_toggle;
  // or this.options = options; // and just access it through options.
},
```
That way it doesn't pollute the actual model object, in the above you would still have access to `this.model` as usual in the expandable view.

js/apps/thirdchannel/views/checkins/date.js
```diff
@@ -0,0 +1,23 @@
...
+        Job = require('thirdchannel/views/checkins/job');
+
+    return Backbone.View.extend({
+        className: 'visit section',
+        template: HandlebarsTemplates['thirdchannel/checkins/date'],
+        render: function() {
+            this.$el.html(this.template(this.model));
+            this.jobs = this.$el.find(".jobs");
+            this.jobs.append("<div>Jobs:</div>");
+            _.each(this.model.job_uuids, function(job_uuid){
```
spullen: Can you loop over the `job_details_by_uuid`?
```javascript
_.each(this.model.job_details_by_uuid, function(job, job_uuid) {
  // ...
});
```

js/apps/thirdchannel/views/checkins/expandable.js
```diff
@@ -0,0 +1,56 @@
...
+            this.$el.html(HandleBarsTemplates['thirdchannel/checkins/expandable']({
+                toggle: this.openText + "&nbsp;<i class='ic ic_down'></i>",
+                hide_toggle: this.model.hide_toggle,
+            }));
+            this.main = this.$el.find(".main");
+            this.main.html(HandleBarsTemplates[this.rowTemplate](this.model));
+            this.subsection = this.$el.find(".subsection");
+            this.toggle = this.$el.find(".toggle");
+            if(this.model.pre_expand){
+               this.trigger("subsection:open");
```
spullen: Why trigger? Can't you just call the method directly?

js/apps/thirdchannel/views/checkins/expandable.js
```diff
@@ -0,0 +1,56 @@
...
+        },
+        initialize: function(){
+            this.listenTo(this, 'subsection:open', this.opensubsection);
+        },
+        render: function() {
+            this.$el.html(HandleBarsTemplates['thirdchannel/checkins/expandable']({
+                toggle: this.openText + "&nbsp;<i class='ic ic_down'></i>",
+                hide_toggle: this.model.hide_toggle,
+            }));
+            this.main = this.$el.find(".main");
```
spullen: Same comment as above, not necessary to do now, but it's just good practice to use the $ prefix when creating variables that represent dom elements.

js/apps/thirdchannel/views/checkins/in_progress.js
```diff
@@ -0,0 +1,19 @@
...
+    var Backbone = require('backbone'),
+        JobView = require('thirdchannel/views/checkins/job');
+
+    return Backbone.View.extend({
+        el: "#jobs",
+        initialize: function(){
+            _.each(this.model.jobs, function(job){
+                var jobView = new JobView({model: {
+                    job: job,
+                    date: "",
```
spullen: empty date?

templates/handlebars/thirdchannel/checkins/date.hbs
```diff
@@ -0,0 +1,22 @@
+<div class="pure-g">
+	{{#if date}}
```
spullen: How about:
```
<div class="date col-4-5 col-md-1-2">
  {{#if date}}
    {{localizedUTCDate date}}
  {{/if}
</div>
```

js/apps/thirdchannel/views/checkins/expandable.js
eliottwiener-thirdchannel: I couldn't figure out a way to.

spullen: `this.opensubsection();` didn't work?

templates/handlebars/thirdchannel/checkins/date.hbs
eliottwiener-thirdchannel: If we don't have a date, `.jobs` is positioned to the left of the task button.

With date:
    <date>    <button>
    <jobs>

Without date:
    <jobs>    <button>

This eliminates the empty space.

spullen: Ah, didn't see the `date` and `jobs` class differentiation. In the else statement you have `col-4-5` and `col-1-2`, the `col-1-2` is that supposed to be `col-md-1-2`

js/apps/thirdchannel/views/checkins/expandable.js
eliottwiener-thirdchannel: I was under the impression that you can't call other view methods from within a view method. I remember having that issue when I was working on a different ticket. It seems to work fine, though.

spullen: If the method was called as a callback then the reference to `this` might be different then the view's `this` which is probably why you ran into issues. You can get around that by adding `_.bindAll(this, 'method1', ...);` in the initialize to all the methods you want to bind the view object to.

js/apps/thirdchannel/views/checkins/date.js
eliottwiener-thirdchannel: Not in this case because `this.model.job_uuids` is a subset of the keys in `this.model.job_details_by_uuid`

spullen: Hmm....
Why are we not just returning a subset of the details list instead of the uuid list and the details list?

eliottwiener-thirdchannel: No real reason. I could do `_.pick` to grab the values from the details list when calling the constructor. In the bootstrap object we have that separation to avoid storing multiple copies of job details (if a job is available for multiple stores/dates).

spullen: Right, I remember talking about this before.

..., I still don't think the duplication really adds too much overhead, but it's fine a suppose.

spullen:
Looks good to me

PR #67, opened 2016-02-02
tc-2816: Don't show contact and address sections unless the user has …

eliottwiener-thirdchannel:
…permission


matthewlittle:
Can a user viewing their own profile see all fields still? If so, then this is good.

PR #68, opened 2016-02-03
tc-1734: UIs for blackout dates

eliottwiener-thirdchannel:
1. Procrastination: When scheduling, there a user should be prevented for scheduling a job on any dates that are blacked out for that job.

2. Oddjob: Management of blackout schemes, and blackout scheme assignments for jobs


spember:
there's a lot going on here, but lgtm :ship:

PR #71, opened 2016-02-11
TC-2898 - displaying the details link for all sales data scenarios

spullen:
https://thirdchannel.atlassian.net/browse/TC-2898

* Adding details link to all sales scenarios as a way to just see the data we have.


neolinkpsu:
lgtm

PR #73, opened 2016-02-22
TC-2889 - adding account similarity list when creating an account

spullen:
https://thirdchannel.atlassian.net/browse/TC-2889

* Display a list of similar account during create
* If there is a direct match, it prevents the form from being submitted
* If there is a partial match, a confirmation dialog is displayed asking to verify that it really isn't already there
* Else it displays a confirm dialog to verify that they want to create the account

Also see PR for Store Service (on bitbucket).

### DIFF
--- a/js/apps/shared/utils/handlebarsHelpers.js
+++ b/js/apps/shared/utils/handlebarsHelpers.js
@@ -227,5 +227,13 @@ define(function (require) {
             timeZoneName: "short"
         });
     });
+
+    Handlebars.registerHelper('displayPercentage', function(value) {
+        if(value === undefined || value === null) {
+            return 'N/A';
+        }
+        value = Math.round(value * 100.0);
+        return value + '%';
+    });
 });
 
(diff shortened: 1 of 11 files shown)


js/apps/shared/utils/handlebarsHelpers.js
```diff
@@ -227,5 +227,21 @@ define(function (require) {
             timeZoneName: "short"
         });
     });
+
+    Handlebars.registerHelper('similarAccountClass', function(similarity) {
+        if(similarity === 1) {
```
spember: Is similarity measured in terms of a float (e.g. 1.0 to 0.0)? Should we show close matches? (e.g. 90% is a close one!)

spember:
See nothing wrong, just one minor question. :ship:

js/apps/shared/utils/handlebarsHelpers.js
spullen: Yes, it is between 1.0 and 0.0. This will show the percentage of those value. They are ordered by similarity descending (done on the backend).

spember: perfect, I thought so. Was just curious as it looks as though we apply a css class on a perfect match, but wondering if adding different classes for close matches would be beneficial. No worries

PR #74, opened 2016-02-23
TC-2169 - swapping the image modal with viewer w/ zoom capabilities

spullen:
Swapped out on:
* full report page
* store profile page (most recent images)
* store image gallery

Also see PR for ThirdChannel

### DIFF
--- a/css/scss/modules/_hoverable_image.scss
+++ b/css/scss/modules/_hoverable_image.scss
@@ -7,21 +7,16 @@
 	cursor: pointer;
 
 	img {
-		max-width: 15em;
-		max-height: 15em;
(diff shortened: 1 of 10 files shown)


neolinkpsu:
seems pretty straight forward.  What browsers/mobile devices have we checked this out on?

spullen:
so far just chrome. I'll take a look at it in some other browsers.

spullen:
Works on firefox and safari. I'll check on a mobile device, less concerned about mobile for these pages though.

spullen:
Tried on a mobile device, seemed to work just fine for me.

PR #75, opened 2016-02-23
tc-2921: build view breakdown links using a merger of info list defau…

eliottwiener-thirdchannel:
…lt filters and the current filters. Always use the filters in the url when loading an info list

https://thirdchannel.atlassian.net/browse/TC-2921
https://github.com/RSV2/ThirdChannel/pull/228

I broke this back when I did https://thirdchannel.atlassian.net/browse/TC-2581

### DIFF
--- a/js/apps/thirdchannel/models/reports/info/info_list.js
+++ b/js/apps/thirdchannel/models/reports/info/info_list.js
@@ -3,13 +3,12 @@ define(function(require) {
 
     return Backbone.Model.extend({
         initialize: function (options) {
-            this.queryString = options.queryString;
             this.programId = options.programId;
             this.reportId = options.reportId;
             this.infoId = options.infoId;
         },
         url: function () {
-            return '/programs/' + this.programId + "/reports/" + this.reportId + "/info/" + this.infoId + ".json?" + this.queryString;
+            return '/programs/' + this.programId + "/reports/" + this.reportId + "/info/" + this.infoId + ".json" + location.search;
         }
     });
 });
--- a/js/apps/thirdchannel/utils/helpers.js
+++ b/js/apps/thirdchannel/utils/helpers.js
@@ -1,5 +1,6 @@
 define(function (require) {
-    var $ = require('jquery');
+    var $ = require('jquery'),
+    deparam = require('jquery-deparam');
     return {
         isMobile: {
             Android: function () {
@@ -55,6 +56,12 @@ define(function (require) {
                     return "Not Mobile";
                 }
             }
+        },
+
+        merge_query_string: function(qs, obj) { // Given a query string and an object
+            var qs_obj = deparam(qs); // turn the query string into an object
+            var merged = _.extend(obj , qs_obj); // merge the two objects with higher precedence for keys in the query string
+            return $.param(merged); // then turn the result back into a query string
         }
     };
 });
--- a/js/apps/thirdchannel/utils/helpers.js
+++ b/js/apps/thirdchannel/utils/helpers.js
@@ -1,6 +1,4 @@
 define(function (require) {
-    var $ = require('jquery'),
-    deparam = require('jquery-deparam');
     return {
         isMobile: {
             Android: function () {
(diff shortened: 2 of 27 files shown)


js/apps/thirdchannel/models/reports/info/info_list.js
```diff
@@ -3,13 +3,12 @@ define(function(require) {
...
     return Backbone.Model.extend({
         initialize: function (options) {
-            this.queryString = options.queryString;
             this.programId = options.programId;
             this.reportId = options.reportId;
             this.infoId = options.infoId;
         },
         url: function () {
-            return '/programs/' + this.programId + "/reports/" + this.reportId + "/info/" + this.infoId + ".json?" + this.queryString;
+            return '/programs/' + this.programId + "/reports/" + this.reportId + "/info/" + this.infoId + ".json" + location.search;
```
spullen: Does `location.search` include the `?`? I'm assuming it does, just want to verify.

js/apps/thirdchannel/views/reports/widgets/bar_chart.js
```diff
@@ -16,8 +17,9 @@ define(function(require) {
             return this;
         },
         updateViewBreakDownLink : function (qs) {
```
spullen: Yikes, we certainly do this function a lot, looks like just one or two that are different. Maybe consider making it a mixin for this? Check this post out http://stackoverflow.com/questions/7853731/proper-way-of-doing-view-mixins-in-backbone

js/apps/thirdchannel/utils/helpers.js
```diff
@@ -55,6 +56,12 @@ define(function (require) {
                     return "Not Mobile";
                 }
             }
+        },
+
+        merge_query_string: function(qs, obj) { // Given a query string and an object
```
spullen: Probably don't need the comments.

js/apps/thirdchannel/views/reports/widgets/totals_averages_table.js
```diff
@@ -2,6 +2,7 @@ define(function(require) {
     var Backbone = require('backbone'),
         Handlebars = require('handlebars'),
         HandlebarsTemplates = require('handlebarsTemplates'),
+        helpers = require('helpers'),
```
spullen: Doesn't looks like this is being used here at the moment, best just to remove.

js/apps/thirdchannel/models/reports/info/info_list.js
eliottwiener-thirdchannel: It includes the `?`.

js/apps/thirdchannel/views/reports/index/widget.js
```diff
@@ -76,61 +78,99 @@ define(function(require) {
             return this;
         },
         createDonutChart: function () {
-            return new DonutChartView(this.model).render().$el;
+            _.extend(DonutChartView.prototype, SharedMixin);
```
spullen: I would apply this where the view is being defined not here.
For example, the DonutChartView defined in thirdchannel/views/reports/widgets/donut_chart:
Instead of doing:
```javascript
return Backbone.View.extend({
// ...
});
```

do
```javascript
var View = Backbone.View.extend({
// ...
});
_.extend(View.prototype, MixinObject);
return View;
```

js/apps/thirdchannel/views/reports/index/widget.js
```diff
@@ -1,8 +1,10 @@
 define(function(require) {
     var Backbone = require('backbone'),
+        _ = require('underscore'),
         Handlebars = require('handlebars'),
         HandlebarsTemplates = require('handlebarsTemplates'),
         HandlebarsHelpers = require('handlebarsHelpers'),
+        SharedMixin = require('thirdchannel/views/reports/widgets/shared_mixin'),
```
spullen: "SharedMixin" is a little vague how about "ViewBreakdownLinkMixin" for both the variable name and the file name.

spullen:
Looks good, just make sure to test on staging before merging.

PR #76, opened 2016-03-11
Tc 3116

neolinkpsu:
Looks like a lot, but really is just a few small things.  Almost all of the hb-compiled files were triggered because I add a loading partial into shared.  (Probably should have been done a long time ago)
The rest is to give a view to the erudition all users page, and the agent profile pages on the opportunities page.


spullen:
Looks fine to me.

PR #77, opened 2016-03-11
tc-2572: add job names to various UI elements

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-2572


spullen:
Looks good to me

PR #79, opened 2016-03-18
TC-3117 - special project dashboard revamp

spullen:
https://thirdchannel.atlassian.net/browse/TC-3117

Also see changes for thirdchannel and oddjob


npm-debug.log
```diff
@@ -0,0 +1,132 @@
+0 info it worked if it ends with ok
```
spember: any particular reason you checked in the npm-debug log ;)?

spember:
aside from the log file, lgtm

npm-debug.log
spullen: Not sure where that came from, i'll remove it.

PR #80, opened 2016-03-21
Added page for the recruiters to get the opportunities urls

neolinkpsu:
Added a simple opportunities page that can be accessed by agents outside of the system
Part 3 of 3 (ThirdChannel, foliage, Erudition)

### DIFF
--- a/templates/handlebars/erudition/recruiting/opportunities/_opportunity.hbs
+++ b/templates/handlebars/erudition/recruiting/opportunities/_opportunity.hbs
@@ -1,0 +1,4 @@
+<div class="body pure-g">
+    <div class="col-1-2 col-md-1">{{name}} {{lastName}}</div>
+    <div class="col-1-2 col-md-1">https://www2.thirdchannel.com/agent/opportunities/{{naturalId}}</div>
+</div>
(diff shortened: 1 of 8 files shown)


templates/handlebars/erudition/recruiting/opportunities/_opportunity.hbs
```diff
@@ -0,0 +1,4 @@
+<div class="body pure-g">
+    <div class="col-1-2 col-md-1">{{name}} {{lastName}}</div>
+    <div class="col-1-2 col-md-1">https://www2.thirdchannel.com/agent/opportunities/{{naturalId}}</div>
```
spullen: Is this supposed to be a link?

neolinkpsu: Nope, this is supposed to be informational.  I figured that they needed an easy way to link their indeed stuff to the new public facing opportunities page for a specific program.  Instead of having them look up the program Id in rails admin and try to construct the url themselves.  I figured this was better for them.   Also I realized the urls was wrong so this comment my get hidden after I update it

templates/handlebars/erudition/recruiting/opportunities/_opportunity.hbs
```diff
@@ -1,4 +1,4 @@
 <div class="body pure-g">
     <div class="col-1-2 col-md-1">{{name}} {{lastName}}</div>
-    <div class="col-1-2 col-md-1">https://www2.thirdchannel.com/agent/opportunities/{{naturalId}}</div>
```
spullen: Follow up, would it make sense to put the generated url from the rails side in the `program.read` response?

neolinkpsu: Actually, good point.  I'll make that change

PR #81, opened 2016-03-22
TC-2888 - allowing the remove button to be displayed for all tasks

spullen:
https://thirdchannel.atlassian.net/browse/TC-2888

* Removes the restriction on the remove link
* Adds *crude* validation to prevent jobs from being submitted without a task associated with it


js/apps/oddjob/views/tasks/create.js
```diff
@@ -36,18 +35,12 @@ define(function (require) {
...
-				this.model.set('removeable', true);
-			}
-
 			return this.model.toJSON();
 		},
 
 		clear: function (e) {
 			e.preventDefault();
-				this.remove();	
+			this.remove();
```
neolinkpsu: not trying to expand the scope of this, but why not just call 'remove' in the event instead of this clear function?

spullen: The edit and create views both use the same functionality. Edit does some additional things that create doesn't.

PR #85, opened 2016-03-28
Added conditions for when the view breakdown should be displayed.  Al…

neolinkpsu:
…so fixed an issue with the list view causing jquery expections.


spullen:
Looks fine to me.

PR #87, opened 2016-04-05
TC-3324 - adding modal to view tags

spullen:
Just adds a way to view the tags that were added to program stores as part of the upload process. This will most likely be expanded on in the future to manage (but as a separate ticket).

Also see PR for store service.


js/apps/stores/templates/hbs-compiled.js
```diff
@@ -1018,7 +1018,7 @@ helpers = this.merge(helpers, Handlebars.helpers); data = data || {};
   if (helper = helpers.name) { stack1 = helper.call(depth0, {hash:{},data:data}); }
   else { helper = (depth0 && depth0.name); stack1 = typeof helper === functionType ? helper.call(depth0, {hash:{},data:data}) : helper; }
   buffer += escapeExpression(stack1)
-    + "</a></td>\n<td>";
+    + "</a></td>\n<td><a href=\"#\" class=\"view-tags\" data-bypass=\"true\">View Tags</a></td>\n<td>";
```
crstalli: Something I found helpful at my last job was coming up with the convention to use ' for javascript strings so that you don't have to escape most of your quotes for attributes.

spullen: This is just a compiled template.

css/scss/apps/stores/_modal.scss
```diff
@@ -1,8 +1,15 @@
-.delete-confirmation-modal {
+.store-modal {
   padding: 1.5em;
+}
 
+.delete-confirmation-modal {
```
neolinkpsu: any reason we removed the 1.5em padding?

neolinkpsu:
👍

css/scss/apps/stores/_modal.scss
spullen: Moved it to the `.store-modal` rule to be re-used for both the store delete and tags list modals.

PR #91, opened 2016-04-11
tc-3233: allowRescheduling view changes

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-3233
https://github.com/RSV2/oddjob/pull/7

### DIFF
--- a/css/scss/apps/oddjob/main.scss
+++ b/css/scss/apps/oddjob/main.scss
@@ -110,6 +110,11 @@
 		margin: 0px 10px 0px 10px !important;
 		padding:10px;
 	}
+	.allow-rescheduling-container {
+		background-color: $light-blue;
+		margin: 0px 10px 0px 10px !important;
+		padding:10px;
+	}
 	.frequencies-container {
 		background-color: $light-blue;
 		margin: 0px 10px 0px 10px !important;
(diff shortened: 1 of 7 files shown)


js/apps/oddjob/views/jobs/edit.js
```diff
@@ -10,6 +10,7 @@ define(function(require) {
 		taskViewClass: require('oddjob/views/tasks/edit'),
 
 		renderChildViews: function () {
+        console.log(this.model);
```
spullen: should remove this.

spullen:
Other than the console log, looks fine to me.

css/scss/apps/oddjob/main.scss
```diff
@@ -110,6 +110,11 @@
 		margin: 0px 10px 0px 10px !important;
 		padding:10px;
 	}
+	.allow-rescheduling-container {
+		background-color: $light-blue;
+		margin: 0px 10px 0px 10px !important;
```
crstalli: I have tried to never use !important. I'd rather reference by id if it is being overridden. Also, you could just seperate with a comma or make the a class more generic class and use the rule above.

PR #92, opened 2016-04-12
Activity feed fixes

neolinkpsu:


css/scss/apps/thirdchannel/_activities.scss
```diff
@@ -29,15 +29,21 @@ $padding: 1.5em;
...
   }
 
+  .task-label {
+    min-width: 5rem;
+  }
+
   .task-item {
-    display: inline-block;
+    display: inline-flex;
+    -webkit-flex-flow: row wrap;
```
spullen: Does this need any other vendored versions?

neolinkpsu: Good point.  Will find and add the other ones

spullen:
Other than the one comment on the vendor css versions, looks good.

PR #96, opened 2016-04-18
Added units sold change to store sales

crstalli:


crstalli:
Tested on staging.

crstalli:
@spullen Could you give this a quick review?

spullen:
Looks fine to me.

In the future could you put a link to the ticket so the reviewer can have a quick way to reference what the issue is. And/or include a description of what was changed and why (especially helpful for more complicated features).

Also, make sure the change gets merged into dev (don't just merge into master).

PR #99, opened 2016-04-27
increase timeout time for reports

crstalli:
Eyeking reports take a long time to generate.


js/apps/thirdchannel/modals/export.js
```diff
@@ -36,7 +36,7 @@ define(function(require){
             var self = this;
             this.xhr = this.model.fetch({
                 success: function (model, response, options) {
-                    if (self.retries++ > 60) {
+                    if (self.retries++ > 120) {
```
spullen: We should make this a configurable option when creating the modal view and default it to 60 if the option is not present.

crstalli:
We actually don't need this yet.

PR #103, opened 2016-05-18
tc-2750: scheduled visits list

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-2750


templates/handlebars/thirdchannel/scheduled_visits/scheduled_visit.hbs
```diff
@@ -0,0 +1,20 @@
+<div class="pure-g scheduled-visit-row">
+	<div class="col-2-12 scheduled-visit-cell">
+		<a href="{{program_store_link}}">
+			{{program_store_account}} {{program_store_tag}}
```
spullen: You should get back a name property on the program store that already does this for you.

spullen:
Just one minor thing, otherwise looks good.

PR #106, opened 2016-05-20
tc-2936: do not track compiled handlebars templates

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-2936


neolinkpsu:
lgtm

PR #108, opened 2016-05-26
Tc 3046

sasha-3C:
This branch incorporates resolutions to both TC-3046 and TC-3540. The other half of this pull request is in the Thirdchannel main app

### DIFF
--- a/js/apps/thirdchannel/views/checkins/show/survey.js
+++ b/js/apps/thirdchannel/views/checkins/show/survey.js
@@ -90,7 +90,9 @@ define(function(require) {
         },
         showElement: function(e) {
             this.$(e.currentTarget.getAttribute('data-show-element')).show('fast', "linear");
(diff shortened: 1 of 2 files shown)


spember:
seems fine to me

js/apps/thirdchannel/views/checkins/show/survey.js
```diff
@@ -90,15 +90,23 @@ define(function(require) {
         },
         showElement: function(e) {
             this.$(e.currentTarget.getAttribute('data-show-element')).show('fast', "linear");
+            if(this.$(e.currentTarget)[0].control.required) {
+                this.$(e.currentTarget.getAttribute('data-show-element')).prop('required', true).trigger('change');
```
spember: this is minor, but I see you're executing $(e.currentTarget.getAttribute('')) twice in this method. It's often more efficient to cache this query in a variable, e.g. var $attribute = this.$(e.currentTarget.getAttribute)), the later call $attribute.show and $attribute.prop

PR #109, opened 2016-05-27
tc-3584: if there is only one job in a visit, pre-expand its tasks au…

eliottwiener-thirdchannel:
…tomatically on the in-progress page

https://thirdchannel.atlassian.net/browse/TC-3584


sasha-3C:
lgtm 👍

PR #111, opened 2016-06-20
Tc 3490

sasha-3C:
Linked PR's:

TC-3490 on oddjob
TC-3490 on Procrastination

This ticket has to do with allowing PMs to remove stores from their respective jobs permanently. When a store closes or no longer needs to be serviced within a job, a PM would like to be able to go into the scheduler and unassign the visit permanently (ie remove the visit from the job itself permanently) rather then have to go through the scheduler, unassign all visits on the agent's side to the store, then go into jobs/tasks and remove the store from all schedules in that job that contain it. The idea was to have all of this taken care of by one button. This button's new home is in the scheduler, right next the "Unassign Visit" button (so you can find it in both the calendar view and the list view).

**CHANGES TO FOLIAGE**
- Added new button to remove stores next to unassign visit button
- Added another parameter to the procrastination call that determines whether we are removing the store or simply just unassigning it

**NOTES**
-There is now a "Remove visit" button located right next to the "Unassign Visit" button in the UI for the scheduler in both places, calendar and list view.
-This button will perform the same actions as the "Unassign visit" button, but for all stores in the job on the scheduler IN ADDITION to removing the store from the job in Jobs/Tasks itself.
-Clicking "Update Existing Schedules" will not cause the store to be added back into the schedule (as it is now removed from Jobs/Tasks for that job).
-Removing a store permanently displays a warning to the user which they must confirm when the button is clicked.
-Though this is probably obvious, removing a store is permanent and thus cannot be reversed - easily, at least.. :)


js/apps/procrastination/templates/hbs-compiled.js
```diff
@@ -0,0 +1,1176 @@
+define(['handlebars'], function(Handlebars) {
```
spullen: Might want to check the .gitignore file. I think hbs-compiled.js files are supposed to be ignored.

spullen:
Just one comment about the hbs-compiled.js file, but other than that looks fine to me.

PR #113, opened 2016-06-21
Adding date ranges to blackout schemes

sasha-3C:
Linked PR: See oddjob TC-3604

This ticket called for allowing users to add a range of dates to blackout schemes vs one at a time. The implementation here was simply adding a range picker and translating the range into a set of dates in the backend. There were no changes made to how the dates were saved and brought back up (if you save a range of dates, when you refresh the page, you'll now see the range of dates translated into a set of individual dates), just a bit of extra translation to deal with the new ranges.  

Updates to foliage:
- added an extra container for new ranges as well as a class for range-pickers (we can have multiple created at a time)
- updated js to send ranges as well as individual dates to oddjob for processing

### DIFF
--- a/css/scss/apps/oddjob/main.scss
+++ b/css/scss/apps/oddjob/main.scss
@@ -279,8 +279,4 @@ div#type-list {
 				width: 100%;
 }
 
-.blackout-scheme-range-row {
-	display: inline-block;
-}
-
 @import "upload";
--- a/css/scss/apps/oddjob/main.scss
+++ b/css/scss/apps/oddjob/main.scss
@@ -279,4 +279,8 @@ div#type-list {
 				width: 100%;
 }
 
+.blackout-scheme-range-row {
+	display: inline-block;
+}
+
 @import "upload";
--- a/js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
+++ b/js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
@@ -67,17 +67,15 @@ define(function (require) {
             }
 
             // create warning for when invalid ranges exist
-            var warningText = "WARNING! You have entered one or more ranges that are invalid and will be disregarded when" +
-            " updating the blackout scheme. \n\nERROR: The date in the 'From' field cannot be after the date in the To field. " +
+            var warningText = "ERROR: You have entered one or more ranges that are invalid." +
+                " \n\nThe date in the 'From' field cannot be after the date in the To field. " +
                 "\n\nInvalid Ranges: ";
 
             _.each(invalidRanges, function(range) {
                 warningText += "\nFrom: " + range.from + ", To: " + range.to;
             });
 
-            warningText += "\n\nContinue anyway?";
-
-            if (invalidRanges.length === 0 || confirm(warningText)) {
+            if (invalidRanges.length === 0) {
                 $("#blackout-scheme-save").prop("disabled", true);
                 var data = {
                     "name": $("#blackout-scheme-name").get(0).value,
@@ -101,6 +99,9 @@ define(function (require) {
                     alert("Failed to save blackout scheme!");
                     $("#blackout-scheme-save").prop("disabled", false);
                 });
+            }
+            else {
+                alert(warningText);
             }
         },
         deleteScheme: function(){
--- a/js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
+++ b/js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
@@ -42,27 +42,62 @@ define(function (require) {
             _.each(window.bootstrap.blackoutScheme.dates, this.addDate);
         },
         submitChanges: function(){
-            $("#blackout-scheme-save").prop("disabled", true);
-            var data = {
-                "name": $("#blackout-scheme-name").get(0).value,
-                "dates": $("#blackout-scheme-dates").find(".blackout-scheme-date-input").map(function(){return this.value;}).get(),
-                "rangesFrom": $("#blackout-scheme-ranges").find(".blackout-scheme-range-input1").map(function(){return this.value;}).get(),
-                "rangesTo": $("#blackout-scheme-ranges").find(".blackout-scheme-range-input2").map(function(){return this.value;}).get(),
-                "inverse": $("#blackout-scheme-inverse_whitelist").is(':checked'),
-            };
-            $.ajax({
-                type: "POST",
-                url: window.bootstrap.submitUrl,
-                data: JSON.stringify(data),
-                success: function(data, textstatus, jqxhr) {
-                    window.location.href = ".";
-                },
-                dataType: 'json',
-                contentType: "application/json; charset=utf-8",
-            }).fail(function(){
-                alert("Failed to save blackout scheme!");
-                $("#blackout-scheme-save").prop("disabled", false);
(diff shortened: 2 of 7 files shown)


templates/handlebars/oddjob/blackout_schemes/range_row.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="blackout-scheme-range-row">
+Date from: <input class="blackout-scheme-range-input1" type="date" required value="{{this}}">
```
eliottwiener-thirdchannel: Capitalization should be consistent: "Date from", "Date To"

eliottwiener-thirdchannel:
The one concern I have here is that there is nothing to stop the user from entering a range with "from" > "to", nor is there a flash message to warn them when they do. This could possibly lead users to think a date range was included when it was not (because the user selected the wrong month or something). I would check with Brendan about that.

Other than that concern, this looks good :+1:

js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
```diff
@@ -18,36 +18,86 @@ define(function (require) {
...
+                }
+            }
+
+            // create warning for when invalid ranges exist
+            var warningText = "WARNING! You have entered one or more ranges that are invalid and will be disregarded when" +
+            " updating the blackout scheme. \n\nERROR: The date in the 'From' field cannot be after the date in the To field. " +
+                "\n\nInvalid Ranges: ";
+
+            _.each(invalidRanges, function(range) {
+                warningText += "\nFrom: " + range.from + ", To: " + range.to;
```
sasha-3C: Added some validation logic here to highlight date ranges that are invalid and show the user a warning if any are found. Will check with Brendan too, but does this new code look alright in general?

js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
```diff
@@ -18,36 +18,86 @@ define(function (require) {
...
+            var rangesFrom = $("#blackout-scheme-ranges").find(".blackout-scheme-range-input1").map(function(){return this.value;}).get();
+            var rangesTo = $("#blackout-scheme-ranges").find(".blackout-scheme-range-input2").map(function(){return this.value;}).get();
+            var invalidRanges = [];
+
+            // validate ranges
+            for(var i = 0; i < rangesFrom.length; i++) {
+                var from = rangesFrom[i];
+                var to = rangesTo[i];
+                var element = $("#blackout-scheme-ranges").find(".blackout-scheme-range-input1").
+                filter(function(){return this.value == from;});
```
eliottwiener-thirdchannel: there is a JSHint error on this. You should define the filter function outside of the loop. Also, move the `.` to the same line as the `filter` call, so it is clear that it is a continuation of the chain on the previous line.

js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
```diff
@@ -18,36 +18,86 @@ define(function (require) {
...
+            " updating the blackout scheme. \n\nERROR: The date in the 'From' field cannot be after the date in the To field. " +
+                "\n\nInvalid Ranges: ";
+
+            _.each(invalidRanges, function(range) {
+                warningText += "\nFrom: " + range.from + ", To: " + range.to;
             });
+
+            warningText += "\n\nContinue anyway?";
+
+            if (invalidRanges.length == 0 || confirm(warningText)) {
```
eliottwiener-thirdchannel: JSHint says to make this `===`

eliottwiener-thirdchannel:
The warning is good, but you should check with Brendan. It may be better to not allow people to submit invalid ranges at all.

In addition, it is usually better to keep the validation logic entirely on the server, since we need it there anyway. In this case, we have checks for range validity both on the client and on the server. If we want to change the validity-checking logic in the future, we will need to update it in both places. The alternative is having the POST endpoint return the results of the server's validity check. For example, it could return something like...

```
{
  "valid": false,
  "reasons": ["Invalid ranges..."],
  "invalid_ranges": [
    { "from": "2016-06-26", "to": "2016-06-22" }
  ]
}
```

...and then the client's error message can be generated based on that response. I would leave it as you have it for now, since it works fine. Just something to keep in mind for later.

:+1:

js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
sasha-3C: Updated

js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
```diff
@@ -18,36 +18,91 @@ define(function (require) {
...
+            var warningText = "ERROR: You have entered one or more ranges that are invalid." +
+                " \n\nThe date in the 'From' field cannot be after the date in the To field. " +
+                "\n\nInvalid Ranges: ";
+
+            _.each(invalidRanges, function(range) {
+                warningText += "\nFrom: " + range.from + ", To: " + range.to;
             });
+
+            if (invalidRanges.length === 0) {
+                $("#blackout-scheme-save").prop("disabled", true);
```
sasha-3C: Talked to Brendan and looks like we want to prevent agents from submitting invalid ranges, so removed the ability to do so. Still kept the warning, just changed it to an alert instead. Does this look good?

templates/handlebars/oddjob/blackout_schemes/range_row.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="blackout-scheme-range-row">
+<div class="blackout-scheme-range-dates" style="display: initial;">
```
eliottwiener-thirdchannel: It's better to be explicit and set this to the `display` value you want instead of using `initial`. In this case you probably want `inline-block` to allow the button to stay on the same line. You could also remove the `blackout-scheme-range-dates` `div` and just have the `input` and `button` elements live together in the `blackout-scheme-range-row`.


We generally also put styling into the stylesheets instead of using the `style` attribute.

js/apps/oddjob/views/blackout_schemes/blackout_scheme.js
eliottwiener-thirdchannel: :+1: looks good

templates/handlebars/oddjob/blackout_schemes/range_row.hbs
sasha-3C: Good point - thanks for the tip 👍

css/scss/apps/oddjob/main.scss
```diff
@@ -279,4 +279,8 @@ div#type-list {
 				width: 100%;
 }
 
+.blackout-scheme-range-row {
```
eliottwiener-thirdchannel: I don't think this is what you want. `inline-block` will cause the `blackout-scheme-range-row`s to gather on one line if the viewport is wide enough. If you want each one to be on its own line, you should use `block`. `block` is the default `display` value for a `div`, so you could probably just leave this out altogether.

https://css-tricks.com/almanac/properties/d/display/

sasha-3C: Oops - yes. You're right - deleted. Sorry, still trying to get the hang of css properties :) thank you

sasha-3C:
Is this set to merge?

eliottwiener-thirdchannel:
yes

PR #114, opened 2016-06-22
Allow user to clone surveys to different customers

sasha-3C:
Linked PR: SingleNickel TC-3005
UI update to allow users to clone surveys to different customers 
- added dropdown to allow users to pick the customer that the survey should clone to when attempting to clone a survey (defaults to current customer).


js/apps/singleNickel/views/survey/list_item.js
```diff
@@ -43,6 +46,20 @@ define(function(require) {
                 context.trigger('error');
             });
         },
+        toggleClone: function(e) {
+            e.preventDefault();
+            var self = this;
+            var container = $("#clone-customer-container");
```
spember: good example of caching a jQuery query

spember:
looks pretty straightforward to me
🚢

PR #119, opened 2016-06-28
Fixing toggle customer bug

sasha-3C:
Relates back to TC-3005: 

Fixed a bug where the clone toggles were only activating on the first survey regardless of which is activated (was an uncaught bug in my own code.. :( )

### DIFF
--- a/js/apps/singleNickel/views/survey/list_item.js
+++ b/js/apps/singleNickel/views/survey/list_item.js
@@ -49,7 +49,7 @@ define(function(require) {
         toggleClone: function(e) {
             e.preventDefault();
             var self = this;
-            var container = $("#clone-customer-container");
+            var container = this.$el.find("#clone-customer-container");
 
             if (container[0].className === 'visible') {
                 container.hide('fast', "linear");
@@ -73,7 +73,7 @@ define(function(require) {
             e.preventDefault();
 
             var self = this;
-            var customerUUID = $("#clone-customer-select")[0].value;
+            var customerUUID = this.$el.find("#clone-customer-select")[0].value;
             this.model.cloneSurvey(null, customerUUID).done(function(response) {
                 if(response.customer_uuid === self.model.get('customer_uuid')) {
                     self.model.collection.add(response);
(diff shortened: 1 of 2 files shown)


eliottwiener-thirdchannel:
`id`s on a page must be unique. Since there are multiple `list_item`s on this page, we should be using `class`es instead for marking the `clone-customer-select` and `clone-customer-container`. Then you can select them with `this.$el.find`.

sasha-3C:
Updated (also added a loading circle that appears when cloning a survey)

js/apps/singleNickel/views/survey/list_item.js
```diff
@@ -48,10 +48,9 @@ define(function(require) {
         },
         toggleClone: function(e) {
             e.preventDefault();
-            var self = this;
-            var container = $("#clone-customer-container");
+            var container = this.$el.find(".clone-customer-container");
 
-            if (container[0].className === 'visible') {
+            if (container[0].className === 'clone-customer-container visible') {
```
eliottwiener-thirdchannel: I would use `container.hasClass('visible')` here to avoid the potentially-brittle string comparison (if somebody adds a class to the `clone-customer-container`, it will break this comparison).

https://api.jquery.com/hasclass/

js/apps/singleNickel/views/survey/list_item.js
```diff
@@ -73,14 +72,18 @@ define(function(require) {
             e.preventDefault();
 
             var self = this;
-            var customerUUID = $("#clone-customer-select")[0].value;
+            var customerUUID = this.$el.find(".clone-customer-select")[0].value;
+            this.$el.find(".clone-customer-container").append("<i class='fa fa-spin fa-spinner'></i>");
```
eliottwiener-thirdchannel: good call

js/apps/singleNickel/views/survey/list_item.js
sasha-3C: good call, fixed

eliottwiener-thirdchannel:
good to merge :+1:

PR #120, opened 2016-06-29
TC-3423 Added size limit checks for file uploads

sasha-3C:
- Added size limit checks for file uploads (cannot be larger than 15 MB) and warnings when the user attempts to upload an image that is too large. 
- Cleaned up the alert logic so that one can specify an alert message to appear (instead of a default 'This may be caused be a bad internet connection!' message, which was showing up when the user for ex. tried to cancel their upload).

### DIFF
--- a/js/apps/shared/views/s3uploader/checkin_file.js
+++ b/js/apps/shared/views/s3uploader/checkin_file.js
@@ -51,7 +51,7 @@ define(function(require) {
             var file = e.target.files[0],
                 self = this;
             
-            if(file && file.size <= 15000000) {
+            if (file && file.size <= (15 << 20)) {
                 var reader = new FileReader();
                 reader.onload = (function(event) {
                     var model = new FileModel({url: self.$form.attr('action'), policy: self.$form.serializeObject(), source: event.target.result});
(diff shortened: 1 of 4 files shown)


js/apps/shared/views/s3uploader/checkin_file.js
```diff
@@ -50,7 +50,8 @@ define(function(require) {
         fileChanged: function(e) {
             var file = e.target.files[0],
                 self = this;
-            if(file !== undefined) {
+            
+            if(file && file.size <= 15000000) {
```
josethirdchannel: you could write this using 15 << 20. It's slightly cleaner and covers the case where a user's phone displays MiB instead of MB.

josethirdchannel:
👍

js/apps/shared/views/s3uploader/checkin_file.js
sasha-3C: So wait..looked this up and << is the shift operator, correct?
Are you suggesting to change this to 
if (file && file.size <= (15 << 20)) 

or do I need to change it in both places?
(unfamiliar with how the operator works)

josethirdchannel: yes, (file && file.size <= (15 << 20)) is equivalent to (file && file.size <= (15 * 2^20)). Not sure what you mean by both places. The line is also in application_file.js which I assume just copies the line from here as part of the grunt task. If not also change it there

PR #121, opened 2016-06-30
TC-2461 Survey export UI

sasha-3C:
Linked PR: SingleNickel TC-2461

These changes were necessary because both Nat and the entirety of the program team are unable to use the yml export that we had before (though I think the yml format works much better for survey data, since both are tree-based...). I updated the UI in foliage so that the user can choose which export file type they'd like to use.


neolinkpsu:
lgtm 👍

PR #122, opened 2016-07-01
TC-4121 Moved clone logic into modal

sasha-3C:
Moving the cloning UI buttons into a new modal. Also added some options that allow the user to navigate directly to the survey or to the new customer after successfully cloning.

### DIFF
--- a/js/apps/singleNickel/modals/clone.js
+++ b/js/apps/singleNickel/modals/clone.js
@@ -1,0 +1,65 @@
+define(function(require){
+    var Backbone = require('backbone'),
+        context = require('context'),
+        BackboneModal = require('backbone.modal'),
+        Handlebars = require('handlebars'),
+        HandlebarsTemplates = require('handlebarsTemplates');
+
+    return Backbone.Modal.extend({
+        templates: {
+            clone: 'singleNickel/modals/clone/clone'
+        },
+        events: {
+            'click .save-clone': 'cloneSurvey',
+            'change .go-to' : 'updateSelect'
+        },
+        cancelEl: '.bbm-button',
+        initialize: function () {
+            this.templateName = this.templates.clone;
+        },
+        template: function () {
+            return HandlebarsTemplates[this.templateName](this.getTemplateData());
+        },
+        updateSelect : function(e) {
+          $(".go-to").not($(e.currentTarget)).prop("checked", false);
+        },
+        cloneSurvey: function(e) {
+            e.preventDefault();
+            var self = this;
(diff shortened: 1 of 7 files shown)


js/apps/singleNickel/modals/clone.js
```diff
@@ -0,0 +1,65 @@
...
+        updateSelect : function(e) {
+          $(".go-to").not($(e.currentTarget)).prop("checked", false);
+        },
+        cloneSurvey: function(e) {
+            e.preventDefault();
+            var self = this;
+            var customerUUID = this.$el.find("#cloneCustomerSelect")[0].value;
+            var goToSurvey = this.$el.find("#goToSurvey")[0].checked;
+            var goToCustomer = this.$el.find("#goToCustomer")[0].checked;
+
```
sasha-3C: Pretty sure I'm supposed to use the ID's here now since we only ever have one modal open at a time? Or should I still use the classes?

eliottwiener-thirdchannel: It's not really a matter of how many instances of a given are visible on screen, but of how many exist in the document. If there are multiple elements with the same id in the document, a `getElementById` will exhibit undefined behavior.

sasha-3C: If the modal is only ever inserted into the document once (in list.hbs), only to be removed/replaced by a different one - does that still count as multiple elements within a document? Not a huge deal to use classes to be safe, I'm just curious. What's an example of a case where we would want to use the ID rather than the class - maybe for the toolbar or something?

eliottwiener-thirdchannel: If the modal replaces the old one, it would be fine to use IDs.

Marking a toolbar with an ID is a good example. IDs are useful for any situation in which you will want to select a specific DOM element. `getElementById` is the fastest type of selector because IDs are (typically) indexed by the browser. This is why jQuery calls `getElementById` directly as a special case when you use a `$('#some-id')` style select:

https://github.com/jquery/jquery/blob/305f193aa57014dc7d8fa0739a3fefd47166cd44/src/core/init.js#L18

sasha-3C: Gotcha - so would using the ID be better in this case since it's faster?

sasha-3C:
Is this good to merge?

js/apps/singleNickel/modals/clone.js
eliottwiener-thirdchannel: It is faster, but the performance difference is negligible in this case.

eliottwiener-thirdchannel:
Yes, looks good.

PR #123, opened 2016-07-18
Tc 3892

josethirdchannel:
https://thirdchannel.atlassian.net/browse/TC-3892


js/apps/thirdchannel/views/comments/new_comment.js
```diff
@@ -25,19 +147,46 @@ define(function(require){
...
                    comment.set(obj);
                    self.collection.add([comment]);
-                   self.$('.new-comment-field').val('');
+                   self.$('.new-comment-field').html('');
                }).fail(function () {
 
                });
            }
        },
+       highlight: function (e) {
```
spullen: Is this being used anymore?

js/apps/thirdchannel/app.js
```diff
@@ -3,11 +3,42 @@ define(function(require){
     var $ = require('jquery'),
         _ = require('underscore'),
         Backbone = require('backbone'),
+        Handlebars = require('handlebars'),
         Router = require('thirdchannel/routers/router');
 
         var initialize = function(){
             Router.initialize();
             Backbone.history.start({pushState: true, hashChange: false});
+            Handlebars.registerHelper("highlight", function (text, mentionedUsers, options) {
```
spullen: I believe we have a shared file that handlebars methods are registered in.

js/apps/thirdchannel/app.js
```diff
@@ -3,11 +3,42 @@ define(function(require){
...
+                var names = [];
+
+                var userNameIdMap = {};
+                for(var i = 0; i < mentionedUsers.length; i++) {
+                    userNameIdMap[mentionedUsers[i].user_name] = mentionedUsers[i].user_id;
+                }
+
+
+                var match;
+                do {
```
spullen: If there are a ton of mentions on the page, how well does it perform. It might be better to throw this stuff in a promise or using rx observables and run this stuff asynchronously.

spullen: it's probably fine.

js/apps/thirdchannel/app.js
eliottwiener-thirdchannel: js/apps/shared/utils/handlebarsHelpers.js

js/apps/thirdchannel/app.js
```diff
@@ -3,11 +3,42 @@ define(function(require){
...
         _ = require('underscore'),
         Backbone = require('backbone'),
+        Handlebars = require('handlebars'),
         Router = require('thirdchannel/routers/router');
 
         var initialize = function(){
             Router.initialize();
             Backbone.history.start({pushState: true, hashChange: false});
+            Handlebars.registerHelper("highlight", function (text, mentionedUsers, options) {
+                var highlightMatcher = RegExp("@(\\w+\\s\\w+)(?:\\sAgents)?\\s\\[[^\\[\\t\\n\\r\\]]+\\]", "g");
```
eliottwiener-thirdchannel: You can get rid of the double-escapes with the alternate regex syntax:

     var highlightMatcher = /yourpatternhere/g

js/apps/thirdchannel/app.js
```diff
@@ -3,11 +3,42 @@ define(function(require){
...
+                    match =  highlightMatcher.exec(text);
+                    if (match) {
+                        mentions.push(match[0]);
+                        names.push(match[1]);
+                    }
+                } while (match);
+
+                for(i = 0; i < mentions.length; i++) {
+                    if (userNameIdMap[names[i]]) {
+                        text = text.replace(mentions[i], "<a class='highlight' href='/programs/Merchandising/profiles/"+userNameIdMap[names[i]]+"'>"+mentions[i]+"</a>");
```
eliottwiener-thirdchannel: You could clean this up by creating the element and setting attributes on it, like:

```
var mentionLink = document.createElement('a')
mentionLink.className = 'highlight';
mentionLink.href = '/programs/Merchandising/profiles' + userNameIdMap[names[i]];
text = text.replace(mentions[i], mentionLink.outerHTML);
```

js/apps/thirdchannel/views/comments/new_comment.js
```diff
@@ -11,11 +11,133 @@ define(function(require){
        initialize: function (options) {
            this.activity = options.activity;
            this.collection = options.collection;
+           $.get('/programs/'+context.programId+'/activities/'+this.activity.id+'/get_mentionable_users', function(data) {
```
eliottwiener-thirdchannel: Might be a good idea to add a `.fail` case to this

josethirdchannel: as of right now a failure just results in the menu not displaying like we saw before. What would be a good failover behavior? log something?

templates/handlebars/thirdchannel/new-comment.hbs
```diff
@@ -2,7 +2,7 @@
 
 <div class="bar-body">
 
-    <textarea rows="1" class="expanding new-comment-field" placeholder="Write a comment..."></textarea>
+    <div contenteditable="true" rows="1" class="expanding new-comment-field" placeholder="Write a comment..."></div>
```
spullen: Why change from textarea to contenteditable? With contenteditable does the rows attribute matter?

js/apps/thirdchannel/views/comments/new_comment.js
```diff
@@ -11,11 +11,132 @@ define(function(require){
...
+                           var text = content.originalText.split(searchMatch);
+                           if (text && text[0] === "") {
+                               text = text.slice(1);
+                           }
+                           var newText = "";
+                           for(var i = 0; i < matches.length; i++) {
+                               newText += "<span class='autocompleteSearchTerm'>"+matches[i]+"</span>" + text[i];
+                           }
+                           content.label = newText;
+                           //content.label = content.originalText.replace(searchMatch,"<span class='autocompleteSearchTerm'>"+searchMatch.source+"</span>")
```
spullen: Dead code, just remove it.

spullen:
Just two additional comments, otherwise looks fine to me.

eliottwiener-thirdchannel:
This is not working for me:

    NoMethodError - undefined method `process_mentions' for #<ActivitiesController:0x007f9b77bb2910>:

Also, I've noticed that if you are matching on a single-character, like with '@a', and then you delete it back to the '@', the autocomplete box will still show the matches for '@a'.

js/apps/thirdchannel/views/comments/new_comment.js
eliottwiener-thirdchannel: We will already have the error logged server-side (unless it's a network issue). It might be good to display an error to the user, so they don't wonder why the @mentions are failing.

eliottwiener-thirdchannel:
I've also noticed that the mention links are not clickable when a comment is first created. They only become clickable after a page reload.

josethirdchannel:
good catch @eliottwiener-thirdchannel. I'll see if I can do something about that

js/apps/shared/utils/handlebarsHelpers.js
```diff
@@ -235,5 +235,43 @@ define(function (require) {
...
+        for(var i = 0; i < mentionedUsers.length; i++) {
+            userNameIdMap[mentionedUsers[i].user_name] = mentionedUsers[i].user_id;
+        }
+
+
+        var match;
+        do {
+            match =  highlightMatcher.exec(text);
+            if (match) {
+                mentions.push(match[0]);
```
spullen: Instead of collecting the mentions, can't you just create the links and what not here?

js/apps/thirdchannel/views/comments/new_comment.js
```diff
@@ -11,11 +11,131 @@ define(function(require){
...
+           $.get('/programs/'+context.programId+'/activities/'+this.activity.id+'/mentionable_users', function(data) {
+               var users = [];
+               var program = data.table.program;
+               if (data.table.can_all_agents) {
+                   users.push({label: "All "+program+" Agents\t[Agent]\t\t", value: "ALL_AGENTS"});
+               }
+               if (data.table.can_all_brand) {
+                   users.push({label: "All "+program+"\t["+program+"]\t\t", value: "ALL_BRAND"});
+               }
+               data.table.users.table.users.forEach(function (item) {
```
spullen: Where is this array being set up, that object chain O_O

josethirdchannel: this comes from activities_controller#mentionable_users. I've removed an unnecessary open_struct to clean it up a little

js/apps/shared/utils/handlebarsHelpers.js
josethirdchannel: since we're modifying the text that highlightMatcher is searching I assumed it might cause some concurrent modification weirdness if there a lot of mentions in a comment.

josethirdchannel:
@eliottwiener-thirdchannel the additional work for mentions should be done in case you want to give everything a last look over

spullen:
looks good to me.

PR #124, opened 2016-07-21
TC-4254: remove function that was adding the discontinued customer_store_id param

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-4254

This removes a function that was adding an `onClick` action to the "View Breakdown" links in order to add a `customer_store_id` param. This was rewriting the `href`s incorrectly, which is why the links broke when opening the links normally. Opening them in a new tab worked fine because it did not trigger the `onClick`. @spember added the function originally, but was not sure what the `customer_store_id` param was used for. It does not seem to be used now, and it doesn't seem like it was ever used in `Reports::InfoController`. In all the cases I tried, it was being set to `null` anyway. The breakdowns appear to work fine without it.


spullen:
Looks fine to me

PR #125, opened 2016-08-02
TC-3431 Opportunities page CSS fixes and adding slidebars to thirdcha…

sasha-3C:
See linked PR on thirdchannel + little-white-lie

Mainly just installing slidebars in thirdchannel and very small CSS fixes

### DIFF
--- a/css/scss/base/_mixins.scss
+++ b/css/scss/base/_mixins.scss
@@ -147,9 +147,9 @@
 }
 
 @mixin swb-medium {
-  @media only screen and (max-width: 48em),
-         only screen and (max-device-width: 48em),
-         only screen and (max-device-width: 48em) and (-webkit-min-device-pixel-ratio: 2) {
+  @media only screen and (max-width: 54em),
+         only screen and (max-device-width: 54em),
+         only screen and (max-device-width: 54em) and (-webkit-min-device-pixel-ratio: 2) {
     @content;
   }
 }
--- a/css/scss/base/_mixins.scss
+++ b/css/scss/base/_mixins.scss
@@ -147,9 +147,9 @@
 }
 
 @mixin swb-medium {
-  @media only screen and (max-width: 54em),
-         only screen and (max-device-width: 54em),
-         only screen and (max-device-width: 54em) and (-webkit-min-device-pixel-ratio: 2) {
+  @media only screen and (max-width: 48em),
+         only screen and (max-device-width: 48em),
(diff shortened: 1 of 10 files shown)


css/scss/base/_mixins.scss
```diff
@@ -147,9 +147,9 @@
 }
 
 @mixin swb-medium {
-  @media only screen and (max-width: 48em),
-         only screen and (max-device-width: 48em),
-         only screen and (max-device-width: 48em) and (-webkit-min-device-pixel-ratio: 2) {
+  @media only screen and (max-width: 54em),
```
neolinkpsu: I won't change this global style unless this has been a global problem, which I don't believe it is.

js/apps/thirdchannel/app.js
```diff
@@ -2,12 +2,14 @@ define(function(require){
...
     var $ = require('jquery'),
         _ = require('underscore'),
+        sb = require('slidebars'),
         Backbone = require('backbone'),
         Router = require('thirdchannel/routers/router');
 
         var initialize = function(){
             Router.initialize();
             Backbone.history.start({pushState: true, hashChange: false});
+            $.slidebars();
```
neolinkpsu: Best not to call it in the main function.  If you need this for certain pages, call it in it's view initialize

css/scss/base/_mixins.scss
sasha-3C: I noticed it was a problem on both the opportunities page and the home page - haven't checked anything else but it looks like it breaks if we ever have the toolbar at the top with the login button. Should I just make a new mixin that has the new constraints and change the broken classes to have that instead?

js/apps/thirdchannel/app.js
sasha-3C: 👍

sasha-3C: So I looked into this and it seems that there isn't a view initialize for the general opportunities page (looks like the front end for that isn't in foliage..?). I need slidebars for the opportunities page and the page for an individual opportunity. Is there a reason the front end for the list of opportunities isn't in foliage but the page for the individual opportunity is?

I can try to rebuild the front end for the page in foliage but not sure if that's worth the effort.

sasha-3C: nvm - decided to just make a small new template with just the initialize

sasha-3C:
Updated

neolinkpsu:
👍

PR #126, opened 2016-08-08
add some leeway with spaces

josethirdchannel:
https://thirdchannel.atlassian.net/browse/TC-4452 
https://thirdchannel.atlassian.net/browse/TC-4479

### DIFF
--- a/js/apps/shared/utils/handlebarsHelpers.js
+++ b/js/apps/shared/utils/handlebarsHelpers.js
@@ -237,7 +237,7 @@ define(function (require) {
     });
 
     Handlebars.registerHelper("highlight", function (text, mentionedUsers, currentUserId, highlightWords, options) {
-        var highlightMatcher = new RegExp(/@(\w+\s{1,3}\w+)(?:\sAgents)?\s\[[^\[\t\n\r\]]+\]/g); // allow for 1-3 spaces between first, last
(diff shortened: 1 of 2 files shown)


js/apps/shared/utils/handlebarsHelpers.js
```diff
@@ -237,7 +237,7 @@ define(function (require) {
     });
 
     Handlebars.registerHelper("highlight", function (text, mentionedUsers, currentUserId, highlightWords, options) {
-        var highlightMatcher = new RegExp(/@(\w+\s{1,3}\w+)(?:\sAgents)?\s\[[^\[\t\n\r\]]+\]/g); // allow for 1-3 spaces between first, last
+        var highlightMatcher = new RegExp(/@(\w+\s{1,3}\w+(?:\sAgents)?\s{1,3})\[[^\[\t\n\r\]]+\]/g); // allow for 1-3 spaces between first, last
```
eliottwiener-thirdchannel: Why did you change the matching group to include the "Agents" part?

josethirdchannel: I added it to cover the case where agents have spaces at the end of their last name. I needed to include the optional (?:Agents)? so I could have a variable number of space characters between between Agent Name and [Role].

eliottwiener-thirdchannel:
:+1:

PR #128, opened 2016-08-11
Tc 4015

sasha-3C:
Disabling start button after clicking to prevent duplicate form submissions

### DIFF
--- a/js/apps/thirdchannel/views/checkins/task_list.js
+++ b/js/apps/thirdchannel/views/checkins/task_list.js
@@ -10,6 +10,7 @@ define(function(require){
         template: HandlebarsTemplates['thirdchannel/checkins/tasks'],
         events: {
             'click .task.grouped button' : 'submit',
+            'click .checkin-form-btn' : 'disableStart',
             "change select": "updateInputs"
         },
         render: function() {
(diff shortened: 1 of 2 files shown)


js/apps/thirdchannel/views/checkins/task_list.js
```diff
@@ -31,6 +32,11 @@ define(function(require){
                 this.showError(e);
             }
         },
+        disableStart: function(e) {
+            e.preventDefault();
+            $(e.currentTarget)[0].disabled = true;
+            this.$('form.checkin-action').submit();
```
spullen: Is `'form.checkin-action'` the same thing as `'.checkin-form-btn'`?

sasha-3C: don't think so - I think 'form.checkin-action' is the whole form, other one is just the submit button

spullen: Yeah, ended up catching it, see updated comment.

sasha-3C: Tested that code and it doesn't seem to work... only disables the button and then doesn't do anything afterwards.

spullen: Fair enough

spullen:
Looks fine to me.

PR #129, opened 2016-08-12
TC-4015 fixed issue with checkin form submissions

sasha-3C:
Disabled start buttons for checkins being submitted

edit: fixed so that the submission selects the correct button


spullen:
looks good to me.

PR #130, opened 2016-08-18
Removed the target logo until we make all the changes necessary for t…

crstalli:
…he honest company

### DIFF
--- a/css/scss/modules/_sales_data_logos.scss
+++ b/css/scss/modules/_sales_data_logos.scss
@@ -7,9 +7,9 @@
 }
 
 .target-logo {
-  display: block;
+  /*display: block;
   width: 250px;
   height: 50px;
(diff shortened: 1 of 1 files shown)


spullen:
Even thought you've commented it out, might be good to clean up the css a little for when we add it back in.

```scss
.sales-logo {
    display: block;
    width: 250px;
    height: 50px;
    background-repeat: no-repeat;

    .sps-logo {
        background-image: url("../images/sps-logo.svg");
     }
 
    .target-logo {
        background-image: url("../images/target-logo.svg");
    }
}
```

crstalli:
We have completely new specs for showing where the data came from. The logos won't be the same size, there will be in a tooltip, not be as big, etc.. It will be completely re-done.

spullen:
Fair enough, then can we just remove the rule altogether then instead of commenting it out.

PR #131, opened 2016-08-19
TC-2299 CSS update for store status on store profile page

sasha-3C:
small CS fixes; see linked PR on Thirdchannel

### DIFF
--- a/css/scss/modules/_address_title.scss
+++ b/css/scss/modules/_address_title.scss
@@ -1,8 +1,25 @@
 .address-title {
+
+	.title-status {
+		margin: 0.9em 0 0.5em 0;
+	}
+
 	.title {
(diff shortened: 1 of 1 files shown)


css/scss/modules/_address_title.scss
```diff
@@ -1,8 +1,25 @@
...
+	.title-status {
+		margin: 0.9em 0 0.5em 0;
+	}
+
 	.title {
 		margin-bottom: 0.5em;
 	}
+
+	.status {
+		font-size: 20px;
```
eliottwiener-thirdchannel: You might want to express this in `em`, so it can scale according to the user's font preference

eliottwiener-thirdchannel:
:+1:

PR #132, opened 2016-08-23
add highlight support for two word display names

josethirdchannel:
https://thirdchannel.atlassian.net/browse/TC-4615


spullen:
Why didn't we just go the easier route of `@AllAgents` for any program?

Looks fine to me.

josethirdchannel:
that looks a lot cleaner. The long form comes from one of the early mockups but I forget what the reasoning behind it was

spullen:
fair enough

👍

PR #133, opened 2016-08-31
Fix for TC-4643

neolinkpsu:


neolinkpsu:
approved by Scott verbally.

PR #134, opened 2016-09-06
TC-4451

josethirdchannel:
https://thirdchannel.atlassian.net/browse/TC-4451

### DIFF
--- a/js/apps/thirdchannel/views/comments/new_comment.js
+++ b/js/apps/thirdchannel/views/comments/new_comment.js
@@ -35,8 +35,9 @@ define(function(require){
                        return;
                    }
                    var splitLabel = ui.item.originalText.split("\t");
-                   var currentText = $(e.target).html();
-                   $(e.target).html(currentText.substring(0, currentText.lastIndexOf('@')+1)+splitLabel[0]+' '+splitLabel[1]);
+                   var $target = $(e.target);
+                   var currentText = $target.html();
+                   $target.html(currentText.substring(0, currentText.lastIndexOf('@')+1)+splitLabel[0]+' '+splitLabel[1]);
 
                    // move cursor to the end
                    var range = document.createRange();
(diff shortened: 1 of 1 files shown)


js/apps/thirdchannel/views/comments/new_comment.js
```diff
@@ -37,6 +37,15 @@ define(function(require){
                    var splitLabel = ui.item.originalText.split("\t");
                    var currentText = $(e.target).html();
                    $(e.target).html(currentText.substring(0, currentText.lastIndexOf('@')+1)+splitLabel[0]+' '+splitLabel[1]);
```
spullen: Because this is being used more than once, might want to cache the jquery object creation

```javascript
var $target = $(e.target);
$target.html(currentText.substring(0, currentText.lastIndexOf('@')+1)+splitLabel[0]+' '+splitLabel[1]);
// trunc ...
var previousMentions = $target.data('mentions');
// trunc ...
$target.data('mentions', mentions);
$target.trigger($.Event("keypress"));
$target.trigger($.Event("change"));
```

js/apps/thirdchannel/views/comments/new_comment.js
```diff
@@ -37,6 +37,15 @@ define(function(require){
...
+
+                   // move cursor to the end
+                   var range = document.createRange();
+                   range.selectNodeContents(e.target);
+                   range.collapse(false);
```
spullen: Is this what ends up getting the cursor to the end of the input? What happens if the user decides to go back and edit somewhere in the middle and add a mention, does it go to the end or after the mention?

josethirdchannel: for now users will only see the mentions dialogue if the type an @ at the end of the text input

js/apps/thirdchannel/views/comments/new_comment.js
josethirdchannel: not sure what you mean `var previousMentions = $target.data('mentions');` is a list of user uuids

spullen: I'm talking about the `$(e.target)`. You can set it to a variable `var $target = $(e.target)` so you don't have to constantly re-create it four additional times in the above script.

js/apps/thirdchannel/views/comments/new_comment.js
spullen: Oh, didn't realize we were just limiting to only appending at the moment, is that by design or a technical limitation?

josethirdchannel: I think it was never really addressed. I implemented it that way for simplicity. It should be possible to insert mentions from anywhere by having the regex check from the user's cursor position instead of the end of the input.

js/apps/thirdchannel/views/comments/new_comment.js
josethirdchannel: I see, I focused on the wrong line. It looks much better now

js/apps/thirdchannel/views/comments/new_comment.js
spullen: well if that's the way it currently works, this is fine

spullen:
👍

PR #136, opened 2016-09-14
Added in new state for special project visits to the gauge (PS-381). …

neolinkpsu:
… Also changed the labeling to be more inline with what we're reporting out which is visits to the store

This is code to add in the new status Visited, but Incomplete. Part 2 of 2 (1 of 2 in ThirdChannel)

### DIFF
--- a/js/apps/thirdchannel/views/dashboards/special_projects/item.js
+++ b/js/apps/thirdchannel/views/dashboards/special_projects/item.js
@@ -20,28 +20,31 @@ define(function(require) {
         },
         renderChart: function() {
             var colors = {
-                Completion: '#96d1b1'
+                Completed: '#96d1b1',
+                'Visited, but incompleted': '#F7B05B'
             };
 
             var data = [];
 
             var percentComplete = this.model.get('percent_of_stores_complete');
+            var percentIncomplete = this.model.get('percent_of_stores_incomplete');
             var target = this.model.get('target');
 
-            if(target) {
-                if(percentComplete < target) {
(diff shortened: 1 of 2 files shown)


js/apps/thirdchannel/views/dashboards/special_projects/item.js
```diff
@@ -20,28 +20,31 @@ define(function(require) {
...
-                }
-            } else {
-                data.push(['Completion', percentComplete]);
+            percentComplete += percentIncomplete;
+
+            data.push(['Target', target]);
+            data.push(['Completed', percentComplete]);
+            data.push(['Visited, but incompleted', percentIncomplete]);
+
+            if(percentComplete <= target) {
```
spullen: I think the reason why I had a guard around target was that some jobs don't specify it (maybe for the legacy case), might want to double check that.

spullen: I think the reason why I had a guard around target was that some jobs don't specify it (maybe for the legacy case), might want to double check that.

spullen:
Just one thing to look into, other than that looks fine to me.

js/apps/thirdchannel/views/dashboards/special_projects/item.js
neolinkpsu: /lib/dashboard_service.rb: line: 141.  We seem to always return a target value.  If it's null it's handled correctly on the front end

spullen: ok

PR #137, opened 2016-09-16
Tc 4755

neolinkpsu:
Issue with the datetimepicker.js library we were using not working correctly.  I've updated the new library and set any instances of that library to the same set of params. Part 2 of 2, first in ThirdChannel


spullen:
Just one question. Other than that seems fine.

js/apps/singleNickel/init.js
```diff
@@ -66,9 +67,6 @@ require.config({
         'chosen': {
             deps: ["jquery"]
         },
-        'dateTimePicker': {
```
neolinkpsu: it's marked as a defines in the library, so no need to do a deps here

PR #140, opened 2016-09-27
Tc 3982

sasha-3C:
Redesigned QTD UI display for activity feed.


css/scss/apps/thirdchannel/_activities.scss
```diff
@@ -149,14 +147,33 @@ $padding: 1.5em;
...
+
+    .sales-button {
+      padding-left: 1em;
+
+      .change-icon {
+        font-size: 1em;
+      }
+    }
+
+    .pure-g {
```
spullen: You may want to just add an extra class just so it doesn't conflict with the pure-g library.

css/scss/apps/thirdchannel/_activities.scss
```diff
@@ -14,11 +14,11 @@ $padding: 1.5em;
   @include clearfix();
   padding: $padding $padding 0 $padding;
   vertical-align: middle;
-  border-bottom: 1px solid $border-color;
+  //border-bottom: 1px solid $border-color;
```
spullen: If we're getting rid of this you can just delete it.

css/scss/apps/thirdchannel/_activities.scss
sasha-3C: Now that I'm looking at it, don't think I even need this class at all. I'll delete the line

spullen: There's an element with `col-2-3` in the container that you removed the `pure-g` class from, are you sure you don't need it?

sasha-3C: I tried deleting it on dev and it didn't seem to break anything. Deleted the col-2-3 thing too and that didn't affect anything either

spullen:
Looks fine to me

PR #144, opened 2016-12-02
TC-2740 Rerendering only task details when switching tasks

sasha-3C:
Fixed an issue where we re-render the whole activity card every time we switch tasks (causing users to lose their comments while they are typing them).
We now instead only re-render the specific task details upon task switch.

**Changes**: created a taskRender() function which the initial render() calls as well as the loadTask() func. Split the activity template up into the task details (which should be re-rendered) and the rest of the activity card.


spullen:
Looks fine to me.

PR #145, opened 2016-12-08
Ps 421

neolinkpsu:
This ticket was to add the sales data which we see on the activity feed to the store list and profile store list page.  This is pull request 2 of 2, first being in thirdchannel

### DIFF
--- a/templates/handlebars/thirdchannel/stores/rows.hbs
+++ b/templates/handlebars/thirdchannel/stores/rows.hbs
@@ -7,17 +7,19 @@
         {{address}}
     </div>
     <div class="col-md-1 col-1-5">
-    {{#if sales_data.message}}
-        <span>
+    {{#if sales_data}}
+        {{#if sales_data.message}}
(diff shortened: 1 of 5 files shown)


templates/handlebars/thirdchannel/stores/rows.hbs
```diff
@@ -1,6 +1,28 @@
...
+<div class="item pure-g">
+    <div class="col-md-1 col-2-5">
+        <a href="{{links.profile}}">{{name}}</a>
+    </div>
+    <div class="col-md-1 col-2-5">
+        {{address}}
+    </div>
+    <div class="col-md-1 col-1-5">
+    {{#if sales_data}}
+        {{#if sales_data.message}}
```
spember: I realize it's likely a limitation of handlebars, but this nested if check (along with the one in /profiles/stores/rows) is a bit weird

spember:
🚢

PR #150, opened 2016-12-28
Trim excess new lines from activity summaries.

aliciasedlock:
**What:** Only ever display a max of two line breaks between paragraphs in activity summaries.

**Why:** Some clients want particularly sectioned summaries from agent. We're already storing summaries with line breaks, so we only had to enable support for them. However, we want to make sure agents aren't abusing line breaks that would cause unnecessarily long entries in the activity feed. For checkins, we replace any instance 4+ line breaks with 3, which adds two lines of space between paragraphs.

**JIRA:** https://thirdchannel.atlassian.net/browse/PS-613

### DIFF
--- a/css/scss/apps/thirdchannel/_activities.scss
+++ b/css/scss/apps/thirdchannel/_activities.scss
@@ -82,7 +82,6 @@ $padding: 1.5em;
 .activity {
   p {
     margin: 0 0 0 0;
-    white-space: pre-wrap;
   }
   background: white;
   border-radius: $border-radius-base;
@@ -331,6 +330,10 @@ $padding: 1.5em;
 
 .activity-content img {
   max-width: 100%;
+}
+
+.activity-content p {
+  white-space: pre-wrap;
 }
 
 .action {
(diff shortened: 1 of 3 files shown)


templates/handlebars/thirdchannel/task.hbs
```diff
@@ -33,7 +33,11 @@
     <div class="summary">
         {{#show_summary}}
             <h3>{{../subject}}</h3>
-            <p>{{{../content}}}</p>
+            {{#if_eq type "Checkin"}}
```
aliciasedlock: We check if it's a checkin since posts get saved at HTML, and we also don't care about the size of the posts for this issue.

spullen:
Looks fine to me

css/scss/apps/thirdchannel/_activities.scss
```diff
@@ -82,7 +82,6 @@ $padding: 1.5em;
 .activity {
   p {
     margin: 0 0 0 0;
-    white-space: pre-wrap;
```
spullen: Was this what caused the weird line breaks for the summary overview (store, person, and datetime) part of an activity?

aliciasedlock: Yep!

spullen:
Looks fine to me.

PR #157, opened 2017-01-10
[SD-258] Calculate sales chart height based on data

aliciasedlock:
**What:** Change how we calculate the height of the sales data graphs to be based on how many items are in the Y axis.

**Why:** These charts were set to a permanent 360px height. Wilson's data does not fit into this height so all the bars were overlapping. We now calculate the height of the chart by seeing if the data would display properly within the 360 limit. If not, we calculate what the minimum needed height is and use that value.

**Jira:** https://thirdchannel.atlassian.net/browse/SD-258

**Before**
screen shot 2017-01-10 at 10 14 23 am

**After**
screen shot 2017-01-11 at 10 26 51 am

### DIFF
--- a/js/apps/thirdchannel/views/store_profile/sales/chart_breakdowns.js
+++ b/js/apps/thirdchannel/views/store_profile/sales/chart_breakdowns.js
@@ -21,6 +21,7 @@ define(function(require) {
         renderSalesPerBreakdown: function() {
             var store = this.model.get('store');
             var breakdowns = [];
+            var graphHeight;
 
             if(this.model.get('breakdown_by') === 'category') {
                 breakdowns = store.categories;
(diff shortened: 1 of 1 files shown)


js/apps/thirdchannel/views/store_profile/sales/chart_breakdowns.js
```diff
@@ -188,11 +194,22 @@ define(function(require) {
...
+
+        getMaxGraphHeight: function(itemLength) {
+          /* 360 = default chart height
+           * 32 = the height of the bar chart (30) + 2 px on either side
+           * for the tick markers that the bars fit between
+           *
+           * This means that when charts get larger than 360px,
+           * the bars will be back to back and not overlapping.
+           */
+          return (itemLength * 32) > 360 ? (itemLength * 32) : 360;
```
crstalli: Do we want to shrink the bars so more fit in the visible area? I would run this by Brendan/Marlene.

aliciasedlock: I'll definitely run it by Marlene, currently the data still does fit into a full window view but I'll get her input on it as well.

I like the `Math.max` use you suggested too, much more straight forward. 👍

aliciasedlock:
Marlene's good with the expanding height of the graph. Her only ask was to provide additional space between the rows, so I've bumped up the allowed space # in our calculation.

PR #158, opened 2017-01-10
[SD-277] Sort sales breakdown table alphabetically

aliciasedlock:
**What:** Sort the sales breakdown chart alphabetically

**Why:** So it's easier to read through.

Within each brand, the categories are already sorted alphabetically, so we only have to sort the top level ourselves.

**Jira:** https://thirdchannel.atlassian.net/browse/SD-277

### DIFF
--- a/js/apps/thirdchannel/views/store_profile/sales/breakdown.js
+++ b/js/apps/thirdchannel/views/store_profile/sales/breakdown.js
@@ -21,7 +21,9 @@ define(function(require) {
 
         renderBreakdownGroups: function() {
             var $body = this.$('.body');
-            this.collection.each(function(breakdownGroup) {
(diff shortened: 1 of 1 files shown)


crstalli:
Code looks good. I would just verify that Oakley works as expected as it only has 1 level breakdown. Also, See if it works correctly when you toggle the setting "Display sales by brand instead of category". The "brand" key is what I am worried about.

crstalli:
If that works then you should be good to go!

aliciasedlock:
@crstalli I looked at this for as many account that I could find sales data for, I think I went between Wilson, Lux, and Oakley. I'll do another round just to make sure, but in comparing their data the use of 'brand' still seemed to represent any of the top level brands or categories. I'll see if I can find an Oakley store with a more in-depth sales breakdown here.

PR #160, opened 2017-01-11
SD-260 - Backend now has Upper Case Categories, so Title cased all the category labels

crstalli:
**What**
The labels of all the brands/categories are now Title Cased. This excludes the bold label in the graph below. 

** Why **
Wilson's Item Master has many different casings. We standardized them on the back end but they are in all caps. That looks ugly. 

Jira: https://thirdchannel.atlassian.net/browse/SD-260

screen shot 2017-01-11 at 1 21 21 pm


aliciasedlock:
The logic itself seems kosher - rather than equate this in three different spots, can we move it into a helper function so it's easier to maintain down the road?

crstalli:
Done

PR #165, opened 2017-01-23
Tc 5008

josethirdchannel:
https://thirdchannel.atlassian.net/browse/TC-5008
This adds an extra check at submission time to make sure a mentioned user is still contained in the final text of the comment

### DIFF
--- a/js/apps/thirdchannel/views/comments/new_comment.js
+++ b/js/apps/thirdchannel/views/comments/new_comment.js
@@ -58,7 +58,7 @@ define(function(require) {
                     selection.addRange(range);
 
                     var mentions = {};
-                    mentions[mention] = ui.item.value;
+                    mentions[ui.item.value] = mention;
                     var previousMentions = $target.data('mentions');
                     if (previousMentions) {
                         mentions = $.extend({}, previousMentions, mentions);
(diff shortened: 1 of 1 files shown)


js/apps/thirdchannel/views/comments/new_comment.js
```diff
@@ -173,10 +174,19 @@ define(function(require) {
             });
 
             if (comment.isValid()) {
+                var mentions = [];
+                var possibleMentions = this.$('.new-comment-field').data('mentions');
+                var text = this.$('.new-comment-field').text();
+                Object.keys(possibleMentions).forEach(function (key) {
+                    if (text.includes(mentions[key])) {
```
eliottwiener-thirdchannel: I think you meant to use `possibleMentions[key]` here

js/apps/thirdchannel/views/comments/new_comment.js
```diff
@@ -46,7 +46,8 @@ define(function(require) {
                     var splitLabel = ui.item.originalText.split("\t");
```
eliottwiener-thirdchannel: It's not immediately obvious what is going on in this `displayItem` function. I had to run through it with some log statements to figure out what was going on. Some comments might make it easier to understand.

josethirdchannel: 👍 agreed

josethirdchannel:
tested by disabling Defenestration and checking the messages sent on the notification topic. It looks like they're still being created as expected but now only include the recipientUUIDs of someone still mentioned in the text on submission

PR #168, opened 2017-01-27
We don't always get the most recent date, so we should display the pa…

crstalli:
…ge even if we dont get it

**What**
Page fails with a javascript error if the store doesn't have sales or if there is no most recent date when you click on the tab. WE should still show the template 

**Why**
We didn't think this wouldn't always be in the response. We will definitely need this when quarters roll over next week.

### DIFF
--- a/js/apps/thirdchannel/views/store_profile/sales/chart_breakdowns.js
+++ b/js/apps/thirdchannel/views/store_profile/sales/chart_breakdowns.js
@@ -4,19 +4,10 @@ define(function(require) {
         context = require('context'),
         c3 = require('c3'),
         Handlebars = require('handlebars'),
+        string_utils = require('thirdchannel/utils/string_utils'),
         HandlebarsTemplates = require('handlebarsTemplates');
 
     var defaultLegendColors = ["#F15F51", "#585E60", "#9FB2C0", "#A9BC4D"];
-
-    var titleizeString = function(text) {
-        if (text) {
-            return text.split(' ').map(function (s) {
-                return s.length <= 1 ? s.toUpperCase() : s.charAt(0).toUpperCase() + s.slice(1).toLowerCase();
-            }).join(" ");
-        } else {
-            return "";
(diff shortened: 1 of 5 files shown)


js/apps/thirdchannel/views/store_profile/sales/chart_breakdowns.js
```diff
@@ -9,9 +9,9 @@ define(function(require) {
     var defaultLegendColors = ["#F15F51", "#585E60", "#9FB2C0", "#A9BC4D"];
 
     var titleizeString = function(text) {
-        return text.split(' ').map(function(s){
+        return text ?  text.split(' ').map(function(s){
```
aliciasedlock: The logic is sound, but from a code organizational standpoint, not a big fan of doing all this logic as part of the return. Not really a blocker but it's a bit easier to digest what's happening now that this is a tad more complicated.

```
var titlizedText = label ? label.split(' ').map(function(s){...

return titlizedText;
```

js/apps/thirdchannel/views/store_profile/sales/report_layout.js
```diff
@@ -139,9 +139,13 @@ define(function(require) {
 
 
             //Title Case the labels
-            return label.split(' ').map(function(s){
-                return s.length <=1 ? s.toUpperCase() : s.charAt(0).toUpperCase() + s.slice(1).toLowerCase();
-            }).join(" ");
+            if (label) {
```
spullen: Can this be moved to a utility function, or if it's something that's being displayed on the page possibly move to a handlebars helper function

crstalli: Alicia and I looked in foliage and there is really no set of utils like this.

spullen: I'm saying you should add one since this is same code is being used in more than one place.

PR #171, opened 2017-02-03
CQ-321 - adding asset tracking problem reports

spullen:
https://thirdchannel.atlassian.net/browse/CQ-321

* Adds views to display the common issues (top 10 issues)
* Adds view to see all issues

Added a filter params model to track the current state of the filter. Needed to do this in order to customize the query strings (mainly for the links for the entity/problem counts link which takes the user to the assets list page filtered to the entity type and problem type). It made it easier to manage that.

Also see PR for indiscipline and ThirdChannel


spember:
pretty straightforward, lgtm

PR #172, opened 2017-02-06
PS-418: more education level options

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/PS-418
https://github.com/RSV2/Erudition/pull/19

Adding the new fields to the profile application and edit forms. The design calls for the education level and graduation year areas to be "greyed out" at times, so I used a new CSS class to accomplish that in a simple way. I also fixed validation for this section to match what we have on the back end.


js/apps/erudition/views/profile/view_profile.js
```diff
@@ -1,51 +1,53 @@
...
 
             return this;
-		},
+        },
 
-		render: function() {
+        render: function() {
             var universityLabel = "";
-            if(this.person.attendedCollege == 'Yes') {
+            if (this.person.attendedCollege == 'Yes') {
```
spullen: Does this need to be update/removed based on the new education statuses?

templates/handlebars/erudition/shared/_university.hbs
```diff
@@ -1,28 +1,46 @@
...
-        </div>
-        <div class="pure-g">
-            <label class="col-1-1">Where did you attend school?</label>
-            <input type="text" class="typeahead" name="university.name" placeholder="Select College/University"
-                   value="{{person.university.name}}" />
-            <input type="hidden" name="university.id" id="university" value="{{person.university.id}}"/>
-        </div>
 
-        <label for="graduationYear">When did you graduate</label>
+				<div id="educationLevelSection">
```
spullen: Can you fix the indentation on this, looks like it's a little out of whack.

js/apps/erudition/views/profile/view_profile.js
spullen: Nvmd on this, I remember what you changed on the backend.

PR #179, opened 2017-02-22
TC-5083: quick wins label management features

eliottwiener-thirdchannel:
companion to
https://bitbucket.org/thirdchannel/storeservice/pull-requests/32/tc-5083-quick-wins-label-management/diff

https://thirdchannel.atlassian.net/browse/TC-5083


templates/handlebars/stores/program_stores/list_item.hbs
```diff
@@ -1,5 +1,5 @@
 <td><a href="/stores/{{storeUUID}}" title="View Store">{{name}}</a></td>
-<td><a href="#" class="view-tags" data-bypass="true" title="{{labels}}">{{labels}}</a></td>
+<td><a href="#" class="view-labels" data-bypass="true" title="{{labels}}">{{labels}}</a></td>
```
spullen: One thing to take into consideration with this is the length of the labels. Since there could be n number of labels, this could cannibalize the entire row (which it why it was put into a modal initially).

eliottwiener-thirdchannel: It will wrap if there is not enough space, so the rows may increase in height. Seems ok, but I will ask Marlene.

PR #181, opened 2017-03-03
Display multiple logos for sales data providers

aliciasedlock:
**What:** Introduces SalesProvidersView, a component that handles displaying multiple logos for sales data providers. Also adds provider logos as part of a new icon font, `sp-icons`.

**Why:** This component can now be used wherever we need to show sales providers fairly easily. The icon font will also lower the work needed if we need to add more provider logos in the future.

**Jira:** https://thirdchannel.atlassian.net/browse/SD-77


css/scss/apps/thirdchannel/_activities.scss
```diff
@@ -16,6 +23,62 @@ $padding: 1.5em;
   }
 }
 
+.sales-data-providers {
+  position: relative;
+  visibility: hidden;
+}
+
+.sales-data-show-providers {
```
crstalli: In a past life I did some f/e and would do something like .sales-data-providers.show and toggle the show class. Just an observation and I have no idea the standard for this ;).

css/scss/apps/thirdchannel/_activities.scss
```diff
@@ -16,6 +23,62 @@ $padding: 1.5em;
...
+    position: absolute;
+    right: 0;
+    text-align: center;
+    top: 3em;
+    width: 13.5em;
+
+    i {
+      display: block;
+      font-size: 4em;
+      margin-left: -1em;
```
crstalli: In said past life I was told never to use negative margins lol.

aliciasedlock: Negative marins aren't always bad, but in the case of the way we're using the font-library, essentially what happens is that all the logos are left aligned, and this is the only way to center them, given that the icons are actually in the `:before`

css/scss/apps/thirdchannel/_activities.scss
aliciasedlock: This isn't being used as a state management class. Can rename the class if it's confusing.

crstalli: No thats fine.

css/scss/apps/thirdchannel/_activities.scss
crstalli: ahhhhh. good point!

PR #183, opened 2017-03-06
TC-5158: fix check for end of activity feed

eliottwiener-thirdchannel:


js/apps/thirdchannel/views/shared/infinite_scroll.js
```diff
@@ -30,7 +30,8 @@ define(function(require) {
                 }
                 this.collection = new this.infiniteCollectionClass({url: this.infiniteURL});
                 this.listenTo(this.collection, 'reset', this.clearAndRender);
-                this.listenTo(this.collection, 'nextPage', this.render);
+                this.listenTo(this.collection, 'nextPage', function(){this.render(false);});
```
spullen: Does the collection class have some state (what page it's on or number of pages), ideally it would be better to use that state to determine whether to fetch more or setting the `allModelsLoaded = true` and `endOfFeed`, it also seems incorrect to pass render any data directly.

eliottwiener-thirdchannel: The page number just increments each time. The collection does not know how many pages there are in total. It just continues until nothing is returned.

spullen: Hmm... Well instead of calling render on the `lastPage` trigger it should just set `allModelsLoaded = true` and call `endOfFeed` and that check can just be removed from the render method altogether. Also, does it make sense for the `endOfFeed` function just to set `allModelsLoaded` as well, would consolidate that.

js/apps/thirdchannel/views/activities/activities.js
```diff
@@ -65,16 +65,6 @@ define(function (require) {
             }
         },
 
-        endOfFeed: function () {
-            // if collection has no models, tell user no activities yet
-            this.loadIndicator.removeFromDOM();
-            if (this.collection.currentPage === 1 && this.collection.models.length === 0) {
```
spullen: Do we not want to display that there are no activities for a given filter set? If this needs to be injected into the parent, maybe you can add variable to the class that stores a message (or a handlebars template) that is conditionally displayed when the current page is 1, the collection length is 0, and the message is set.

js/apps/thirdchannel/views/shared/infinite_scroll.js
```diff
@@ -31,6 +31,7 @@ define(function(require) {
                 this.collection = new this.infiniteCollectionClass({url: this.infiniteURL});
                 this.listenTo(this.collection, 'reset', this.clearAndRender);
                 this.listenTo(this.collection, 'nextPage', this.render);
+                this.listenTo(this.collection, 'reachedEnd', this.endOfFeed);
```
spullen: This looks much cleaner!

PR #186, opened 2017-03-16
TC-5199: don't load the cost estimate views for vega agents

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/TC-5199

This is intentionally a "quick and dirty" way of doing this.

### DIFF
--- a/js/apps/procrastination/routers/main.js
+++ b/js/apps/procrastination/routers/main.js
@@ -42,8 +42,10 @@ define(function(require) {
 
             createSchedule: function(customerSlug, programSlug, personId, aggregateId) {
                 new SetSchedule({aggregateId: aggregateId});
-                var model = new CostEstimate({id: aggregateId});
-                new CostEstimateView({model: model}).fetch();
+                if(programSlug !== 'vega'){
+                    var model = new CostEstimate({id: aggregateId});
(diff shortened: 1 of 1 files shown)


tasks/procrastination/jshint.js
```diff
@@ -11,11 +11,12 @@ module.exports = function(grunt) {
         // the all task covers all files, excluding the hbs-compiled (auto-generated) and any libs we use (we didn't write them)
         procrastination: ['js/apps/procrastination/**/*.js'],
         options: {
+            debug: true,
```
aliciasedlock: Did you intend to leave this in?

templates/handlebars/procrastination/admin/cost_estimate.hbs
```diff
@@ -1 +1 @@
-<span>Estimated zone rate: ${{travel}}</span>
\ No newline at end of file
+<span>admin Estimated zone rate: ${{travel}}</span>
```
aliciasedlock: Is this and the other `cost_estimate` template supposed to have this copy update? If so the capitalization seems weird.

eliottwiener-thirdchannel: oops I accidentally did a `-am` instead of `-m`

PR #194, opened 2017-04-13
VEGA-24: show unschedule button only for incomplete visits and for pe…

eliottwiener-thirdchannel:
…ople that the backend say can unschedule

https://thirdchannel.atlassian.net/browse/VEGA-24

### DIFF
--- a/js/apps/procrastination/models/schedule/main.js
+++ b/js/apps/procrastination/models/schedule/main.js
@@ -11,6 +11,7 @@ define(function (require) {
                 }
                 if(!this.get('dateCompleted')) {
                     this.set({canUnassign: context.canUnassign});
+                    this.set({canUnschedule: context.canUnschedule});
                 }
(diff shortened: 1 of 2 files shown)


js/apps/procrastination/models/schedule/main.js
```diff
@@ -11,6 +11,7 @@ define(function (require) {
                 }
                 if(!this.get('dateCompleted')) {
                     this.set({canUnassign: context.canUnassign});
```
josethirdchannel: I'm assuming there's a default value for this somewhere? Is canUnassign and canUnschedule set to false by default and only explicitly allowed for admins(assuming admins means PMs and not Vega personnel).

eliottwiener-thirdchannel: It's set by the backend. True only for admins.

PR #207, opened 2017-04-26
PS-910: Stores export on profile stores page

katelynsalvatori:


js/apps/thirdchannel/models/profiles/stores_export.js
```diff
@@ -0,0 +1,14 @@
+define(function(require) {
+    var Backbone = require('backbone');
+
+    return Backbone.Model.extend({
+        initialize: function (options) {
+            console.log(options.programId);
```
neolinkpsu: Should clean these up

PR #208, opened 2017-04-27
[TC-5063] Add full-width mode to SalesMarquee, add to checkin workflow

aliciasedlock:
**Why:** We want FMRs to be able to see stores sales data when completing their surveys.

**Jira:** https://thirdchannel.atlassian.net/browse/TC-5063

### DIFF
--- a/js/apps/thirdchannel/views/shared/sales_marquee.js
+++ b/js/apps/thirdchannel/views/shared/sales_marquee.js
@@ -18,6 +18,7 @@ define(function(require) {
         initialize: function (data) {
           this.currentWidget = 0;
           this.salesData = data.salesData;
+          this.display = data.display;
           this.salesWidgets = this.formatSalesWidgets(this.salesData);
           this.canNavBack = false;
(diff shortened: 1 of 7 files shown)


js/apps/thirdchannel/views/shared/sales_marquee.js
```diff
@@ -26,29 +27,32 @@ define(function(require) {
         },
 
         render: function () {
+          var templateData;
+
           this.canNavBack = (this.salesWidgets.length !== 0 && this.currentWidget !== 0);
           this.canNavForward = (this.salesWidgets.length !== 0 && this.currentWidget !== this.salesWidgets.length - 1);
 
-          var templateData = {
+          templateData = {
```
josethirdchannel: was losing this var intentional?

aliciasedlock: Oops, nope, this was missed from undoing a change I thought I needed. Will fix!

PR #211, opened 2017-04-28
[FAS-105] On initial load, match slider state to current filter state

aliciasedlock:
**What:** When the date slider initializes, check the page's current filters, and compare to the slider's current date mapping. If there's a match in ranges, set the slider position to that range, otherwise, default to the "Custom" option.

**Why:** When navigating back and forth between this page, the slider would also reset to 1 month, even if the filters were different in the URL. This also allows us to better deeplink this page so we can send URLs with particular dates filtered by default.

**Jira:** https://thirdchannel.atlassian.net/browse/FAS-105

### DIFF
--- a/js/apps/thirdchannel/views/filter/dateSlider.js
+++ b/js/apps/thirdchannel/views/filter/dateSlider.js
@@ -117,22 +117,22 @@ define(function(require) {
         },
 
         setStartPointFromFilters: function() {
+          var startPoint = Object.keys(this.dateMap).length - 1; // Set a default startPoint to last item in dateMap
           var startDateFilter = _.find(this.filters.components, function(filter) {
             return filter.filterParam === "start_date";
-          }).activeFilters[0].options.value;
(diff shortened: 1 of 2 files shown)


js/apps/thirdchannel/views/filter/dateSlider.js
```diff
@@ -114,6 +116,25 @@ define(function(require) {
           context.trigger('filter:set', filterUpdates);
         },
 
+        setStartPointFromFilters: function() {
+          var startPoint = Object.keys(this.dateMap).length - 1; // Set a default startPoint to last item in dateMap
+          var startDateFilter = _.find(this.filters.components, function(filter) {
+            return filter.filterParam === "start_date";
+          }).activeFilters[0].getQueryValue(); // start_date should only ever have a max of one active item
```
aliciasedlock: Is there a better way to get this single value? I couldn't really find a good way but may not have looked in the right place for it.

PR #236, opened 2017-06-05
FAS-203: Add support for multiple KPIs footnotes

mmichael0413:
https://thirdchannel.atlassian.net/browse/FAS-203


templates/handlebars/thirdchannel/reports/field_activity/kpis/average_and_top_3.hbs
```diff
@@ -2,36 +2,31 @@
...
+          <ol>
+          {{#each data.1.value}}
+              <li>{{this}}</li>
+          {{/each}}
+          </ol>
+          {{data.1.postfix}}
+      </span>
+    </div>
+
+    {{#if footnote}}
```
neolinkpsu: Minor things, but I believe you could place the if into the footnote partial, so that you don't have to do this everywhere

PR #242, opened 2017-06-14
VEGA-173: confirmed start time

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/VEGA-173
https://github.com/RSV2/Procrastination/pull/88
https://github.com/RSV2/ThirdChannel/pull/674

For procrastination: adds the timepicker to the confirmation form, and shows the time in the start time column

For thirdchannel: shows confirmed start time in its own column on the main job requests list


js/apps/procrastination/views/schedule/current/list_visit.js
```diff
@@ -89,12 +96,14 @@ define(function (require) {
         confirmVisit: function (e) {
             e.preventDefault();
             e.stopPropagation();
-            var url = context.base_url+"/"+ this.model.get('personId')+"/show/"+this.model.get('aggregateId')+"/"+this.model.get('id')+"/confirm";
+            confirmedStartTime = this.$('.start_time').val();
+            var url = context.base_url+"/"+ this.model.get('personId')+"/show/"+this.model.get('aggregateId')+"/"+this.model.get('id')+"/confirm?confirmedStartTime="+confirmedStartTime;
```
spullen: Since this is a post, would it be better to pass `confirmedStartTime` as data instead of as a query string?

PR #243, opened 2017-06-19
PS-1055: Transfer Social Handles to profile instead of application

mmichael0413:
https://thirdchannel.atlassian.net/browse/PS-1055


js/apps/erudition/views/profile/edit.js
```diff
@@ -93,14 +93,48 @@ define(function (require) {
...
+                    }
+                }
+
+                return valid;
+            }, 'Invalid Instagram handle.');
+
+            $.validator.addMethod('validateFacebookUrl', function(value, element) {
+                var valid = false,
+                    url = self.$('input[name=facebookUrl]').val(),
+                    regex = /^(https?:\/\/)?((w{3}\.)?)facebook.com\/.*/;
```
josethirdchannel: we might also want to validate on length here given that the column in erudition is limited to 255 characters. Also, are we trimming the url params e.g. everything after the `?`

mmichael0413: The length validation is actually in the handlebars template now.

mmichael0413:
Also removed the `data-rule-required` attribute for Street Address 2.

js/apps/erudition/views/profile/edit.js
```diff
@@ -121,6 +122,54 @@ define(function (require) {
...
+
+                return valid;
+            }, 'Invalid Instagram handle.');
+
+            $.validator.addMethod('validateFacebookUrl', function(value, element) {
+                var valid = false,
+                    url = self.$('input[name=facebookUrl]').val(),
+                    regex = /^(https?:\/\/)((w{3}\.)?)facebook.com\/.*/;
+
+                if (url.length > 0) {
```
spullen: Could be a bit more defensive. `if(url && url.length > 0 && regex.exec(url))`

js/apps/erudition/views/profile/edit.js
```diff
@@ -121,6 +122,54 @@ define(function (require) {
...
+                    if (regex.exec(url)) {
+                        valid = true;
+                    }
+                } else {
+                    valid = true;
```
spullen: Should this be `false`?

js/apps/erudition/views/profile/edit.js
```diff
@@ -121,6 +122,54 @@ define(function (require) {
             });
         },
 
+        validateSocialMediaProfiles: function() {
+            $.validator.addMethod('validateInstagramHandle', function(value, element) {
+                var valid = false,
+                    handle = self.$('input[name=instagramHandle]').val(),
+                    regex = /^@[a-zA-Z0-9\.\-_]{0,28}$/;
+
+                if (handle.length > 0) {
```
spullen: Could be a bit more defensive here, and combine the conditionals

js/apps/erudition/views/profile/edit.js
```diff
@@ -121,6 +122,54 @@ define(function (require) {
...
+                    if (regex.exec(handle)) {
+                        valid = true;
+                    }
+                } else {
+                    valid = true;
```
spullen: Same, should this be `false`

mmichael0413: The default is false so that's setting it to true if the value is empty (since these fields aren't required).

spullen: I see, the way you have all of these conditionals and branch make it a little difficult to read.

Maybe:
```javascript
if(handle === null || handle === "" || (handle.length > 0 && regex.exec(handle)) {
    return true;
} else {
    return false;
}
```

Or could probably be condensed to just

```javascript
if(handle === "" || regex.exec(handle)) {
    return true;
} else {
    return false;
}
```
depending on how defensive you want to be. 

Same goes for the other validation below.

mmichael0413: Agreed. I was thinking:

```
if (url && ((url.length > 0 && regex.exec(url)) || url.length === 0)) {
  valid = true;
}
```

mmichael0413: I do like the explicit return, so maybe combine the condensed conditionals with the false/true return instead of returning `valid`?

spullen: You'll need to check that it's not there first

```javascript
if(!url || url.length === 0 || regex.exec(url)) {
    return true;
} else {
    return false;
}
```
That way you don't need to check the length is greater than zero, because it's implied by checking that it's zero first. This also reads better, "if the url/handle is not present or is a string of length zero or is present and has the correct format return true, otherwise return false"

js/apps/erudition/views/profile/edit.js
```diff
@@ -121,6 +122,54 @@ define(function (require) {
             });
         },
 
+        validateSocialMediaProfiles: function() {
+            $.validator.addMethod('validateInstagramHandle', function(value, element) {
+                var valid = false,
+                    handle = self.$('input[name=instagramHandle]').val(),
+                    regex = /^@[a-zA-Z0-9\.\-_]{0,28}$/;
```
spullen: Can make these regular expressions constant so they're not created each time.

js/apps/erudition/views/profile/edit.js
```diff
@@ -121,6 +122,54 @@ define(function (require) {
             });
         },
 
+        validateSocialMediaProfiles: function() {
+            $.validator.addMethod('validateInstagramHandle', function(value, element) {
+                var valid = false,
+                    handle = self.$('input[name=instagramHandle]').val(),
```
spullen: Do you need to do this if the value is passed into the function already?

js/apps/erudition/views/profile/edit.js
```diff
@@ -121,6 +122,42 @@ define(function (require) {
             });
         },
 
+        validateSocialMediaProfiles: function() {
+            $.validator.addMethod('validateInstagramHandle', function(value, element) {
+                var INSTAGRAM_REGEX = /^@[a-zA-Z0-9\.\-_]{0,28}$/;
```
mmichael0413: `const` isn't supported by our JS version.

spullen: `const` doesn't mean it's a constant, just means it can't be changed to another value once it's been set (read https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75). And it's supported by some browser versions, but not all atm (give it a try in your chrome console). For this one you can just set it outside of the backbone view definition (under the `var $ = require('jquery'), ....` requires;

So:
```javascript
define(function (require) {
    require('jquery-ui');
    require('jquery.mask');
    require('chosen');
    require('jquery-validate');

    var $ = require('jquery'),
        context = require('context'),
        templates = require('handlebarsTemplates'),
        typeahead = require('typeahead'),
        Chosen = require('chosen'),
        FileView = require('shared/views/s3uploader/application_file');

    var INSTAGRAM_REGEX = /^@[a-zA-Z0-9\.\-_]{0,28}$/,
          FACEBOOK_REGEX = /^(https?:\/\/)((w{3}\.)?)facebook.com\/.*/;

    return Backbone.View.extend({
        // ... the rest
    });
});
```

mmichael0413: The linter fails when I use `const`. Pretty sure most browsers support it though so we could update `jshint` to use ES6 at some point.

PR #261, opened 2017-07-19
SCHED-45: set default selection to 12:00pm

eliottwiener-thirdchannel:
https://thirdchannel.atlassian.net/browse/sched-45

An existing start time was already being selected, as desired. I changed the default start time selection to 12:00pm.


templates/handlebars/procrastination/schedule/schedule_row.hbs
```diff
@@ -35,7 +35,11 @@
         {{ startTime }}
       {{/if}}
     {{else}}
-      <input type="text" name="start_time" class="start_time" value="{{startTime}}" />
+      {{#if startTime}}
+        <input type="text" name="start_time" class="start_time" value="{{startTime}}" />
+      {{ else }}
+        <input type="text" name="start_time" class="start_time" value="12:00pm" />
```
josethirdchannel: is this really the format the backend expects?

templates/handlebars/procrastination/schedule/schedule_row.hbs
```diff
@@ -35,7 +35,11 @@
         {{ startTime }}
       {{/if}}
     {{else}}
-      <input type="text" name="start_time" class="start_time" value="{{startTime}}" />
+      {{#if startTime}}
```
spullen: Been trying to be better about stateless templates. With that said instead of conditionally displaying the same input why not determine the value in the view object that gets passed to the template.

example:
view:
```javascript
Backbone.View.extend({
  // ... stuff
  render: function() {
    var data = {}
    data.selectedStartTime = this.model.get('startTime') === null ? "12:00pm" : this.model.get('startTime');
    this.$el.html(template(data));
    return this;
  }
  // more stuff...
});
```
template:
```javascript
<input type="text" name="start_time" class="start_time" value="{{selectedStartTime}}" />
```

templates/handlebars/procrastination/schedule/schedule_row.hbs
eliottwiener-thirdchannel: yes we just store start time as a string with that format

PR #262, opened 2017-07-19
Vega 216

josethirdchannel:
https://thirdchannel.atlassian.net/browse/VEGA-216


spullen:
One other thing that I didn't realize when I first looked through this. The scheduled visits edit page (the calendar view) now shows the expiration date. For this ticket, if the direct assignee expiration date is set, we should show that instead of the normal expiration date.

templates/handlebars/procrastination/schedule/upcoming/visit.hbs
```diff
@@ -18,7 +17,7 @@
     </div>
 
     {{> 'tasks'}}
-
+    </div>
```
spullen: So looks like you fixed one of the closing divs, but added this one. Remove this one, and add a closing div tag after line 10 (`<div>{{address}}</div>`)

templates/handlebars/procrastination/schedule/upcoming/visit.hbs
```diff
@@ -29,8 +28,12 @@
             <a href="#" class='remove btn light small'>Remove this store from this job</a>
         {{/if}}
 
-        {{#if jobDetails.expirationDate}}
-            <span class="expiration-date">Expires: {{dateFormat jobDetails.expirationDate 'MMM D, YYYY'}}</span>
+        {{#if jobDetails.directAssigneeExpirationDate}}
```
spullen: So once the job has expired from this condition, we need to show the real expiration date. This might be something that can be determined in view or even on the backend. Ideally we should just have to have one `<span class="expiration-date">Expires: {{dateFormat jobDetails.expirationDate 'MMM D, YYYY'}}</span>`s

templates/handlebars/thirdchannel/manage/jobs/create.hbs
```diff
@@ -85,6 +85,13 @@
 
         {{#if canConfigureOpenJobRequest}}
             <div class="form-group">
+                <label>Direct Assignee Expiration Date</label>
+                <input type="text" name="direct_assignee_expiration_date" class="direct_assignee_expiration_date" value="{{direct_assignee_expiration_date}}" />
+            </div>
+        {{/if}}
```
spullen: Since we have two components that relying on the `canConfigureOpenJobRequest` that follow each other you can remove the `{{/if]}` here on line 91 and the `{{#if canConfigureOpenJobRequest}}` on line 93.

templates/handlebars/procrastination/schedule/upcoming/visit.hbs
josethirdchannel: Right now I have the consumer in the Scheduling service clear out `directAssigneeExpirationDate` once a job expires as a way to mark these jobs and avoid clearing out someone who has accepted the job after the initial assignee expiration

spullen:
Also realized from looking at the code on thirdchannel. We probably only want to display the direct assignee expiration date if an assignee has been selected.

PR #263, opened 2017-07-20
replace job request surveys with trackables

mmichael0413:


templates/handlebars/thirdchannel/manage/jobs/create.hbs
```diff
@@ -10,10 +10,12 @@
 
         <div class="form-group">
             <label>Visit Type</label>
-            <select class="survey_uuid" name="survey_uuid" required="required">
-                {{#select survey_uuid}}
-                    {{#each surveys}}
-                        <option name="survey_uuid" class="survey_uuid" value="{{this.uuid}}">{{capitalize this.survey_type}}</option>
+            <select class="trackable_id" name="trackable_id" required="required">
+                {{#select id}}
```
spullen: Will this work, shouldn't this be `trackable_id`? `id` is the id of the job request.

mmichael0413: Yes, the `id` is just the uuid of the trackable object. These get set in `jobs#build_bootstrap`:

```
trackables: trackables.map{ |t| { id: t.uuid, name: t.full_name } }
```

mmichael0413: Good catch!

PR #275, opened 2017-08-09
SCHED-65 - updating views related to date confirmed and confirmation start time

spullen:
https://thirdchannel.atlassian.net/browse/SCHED-65

Also see https://github.com/thirdchannel/job-scheduling/pull/115


templates/handlebars/procrastination/schedule/schedule_row.hbs
```diff
@@ -15,33 +15,31 @@
 {{! Duration }}
 <div class="col-2-12 hidden-xs">{{totalDuration}} mins</div>
 
-{{#if canConfirm}}
```
eliottwiener-thirdchannel: Isn't this check there to hide this column for non-job-request programs?

spullen: Ah, true, good catch.

PR #277, opened 2017-08-16
PS-1216: Added activity actions and commenting to full report

katelynsalvatori:
Also changed following button text for PS-1229


js/apps/thirdchannel/views/activities/activity.js
```diff
@@ -260,6 +156,18 @@ define(function(require) {
                 new SalesMarquee({el: this.$el.find('.sales-marquee'), salesData: event});
                 this.$el.find('.sales-widget').fadeIn(500).css("display","inline-block");
             }
+        },
+        followActivity: function(e) {
+            var self = this;
+            self.activityActions.followActivity(e);
+            setTimeout(function() {
```
mmichael0413: Not sure if using a timeout here is the right call, since wrapping `self.activityActions.followActivity(e)` with a promise returned an error, and I need the `followActivity` function to finish executing before getting the updated `follow_count` (since that function updates the value in the state, from which I then read).

PR #285, opened 2017-09-05
WF-116 - displaying duration in hours, unless under an hour for scheduling system

spullen:
https://thirdchannel.atlassian.net/browse/WF-116


spullen:
I think I accidentally branched off of that branch. But should be fine since I'm planning on releasing that tomorrow anyways and still have to test this on staging.

PR #293, opened 2017-09-21
Wf 119

josethirdchannel:
https://thirdchannel.atlassian.net/browse/WF-119

individual ranges all have to be within the same month . The calendar should grey out days when selecting the end date based on the begin date.

at save time we'll do a check across all ranges to make sure they're in the same month


js/apps/thirdchannel/views/manage/jobs/dateRange.js
```diff
@@ -45,6 +45,7 @@ define(function(require) {
                 startDate: startDate,
                 minDate: startDate,
                 onShow: function(ct) {
+                    self.$('.end').val(null);
```
spullen: This doesn't seem right, if we null the value then get the value to determine things to display on show, it'll always be null. Unless I'm missing something?

spullen: If it's always supposed to show the maxDate, if set, then this logic for end and maxDate here should be removed.

josethirdchannel: agreed, it really should just clear out the end date & always show maxDate

spullen: I don't believe you want to null out the value at all! Say user selected value, closed the selector, then opened selector without changing anything, and then closed it. Wouldn't that null out the value the selected?

spullen: I think this should just be left as is (prior to the change), all it's doing is saying if someone selected the end date, the start should be allowed after the end (or max date)

js/apps/thirdchannel/views/manage/jobs/dateRange.js
```diff
@@ -44,11 +44,9 @@ define(function(require) {
                 scrollInput: false,
                 startDate: startDate,
                 minDate: startDate,
+                maxDate: maxDate,
                 onShow: function(ct) {
```
spullen: Just copying my comment since it was hidden:

I think this should just be left as is (prior to the change), all it's doing is saying if someone selected the end date, the start shouldn't be allowed after the end (or max date)

josethirdchannel: now that we're restricting dates to be within the same month having the old logic makes it impossible for users to change the month for an existing date range. They'd have to delete it and start over. That seems like it would make editing more cumbersome, though maybe that's desirable?

opening the selector without changing anything does clear out the end date. We could use :onChangeDateTime instead of :onShow to fix that

spullen: I would play around with this. Try this, first select an end date of October 10, then select a start date in September.

josethirdchannel: The code as written right now would 

select Oct. 10 for the end date
clear out the end date when the user edits the start date
allow the user to select any start date in any month up until max date

The alternative though would be to 

select Oct 10 as the end date 
only allow the user to select a start date in October
then that range cannot be edited to any date outside of October

the code as is is somewhat prescriptive in that it basically forces you to enter date ranges start -> end. Though option 2 also seems clunky

PR #299, opened 2017-10-03
WF-61 - batch status update for visit requests

spullen:
https://thirdchannel.atlassian.net/browse/WF-61

Displays which statuses to transition to based on what was selected. See ticket for details on those transitions. Outside the ticket, added the ability to transition from confirmed to accepted to force people to reconfirm since we already have that functionality for singular visit requests, this seemed like a logical thing to add.

Also, note that the ticket has information about batch reset decline, but that has been separated into it's own ticket.

Also PRs for thirdchannel and scheduling


js/apps/thirdchannel/views/manage/jobs/list.js
```diff
@@ -198,11 +215,53 @@ define(function(require) {
...
+                    'expired': 0
+                }, statusCounts);
+
+                // Determine which status change options are enabled/disabled.
+                // Currently it's all or nothing. All the selected visit requests must be able to transition to the same status
+
+                // not_published, withdrawn, expired, accepted, or confirmed to allow to transition to published
+                var publishableCount = statusCounts[NOT_PUBLISHED] + statusCounts[WITHDRAWN] + statusCounts[EXPIRED] +
+                    statusCounts[ACCEPTED] + statusCounts[CONFIRMED];
+
```
josethirdchannel: we could probably include published in here to prevent selecting a published visit from blocking other operations even though it's effectively a no-op

spullen: That's a good idea, there are cases where we just want to send an update to fix something, where that would be super convenient

PR #303, opened 2017-10-11
open the calendar to the selected start date

josethirdchannel:
https://thirdchannel.atlassian.net/browse/WF-185

small fix to have the calendar default to the current date when editing


js/apps/thirdchannel/views/manage/jobs/dateRange.js
```diff
@@ -46,7 +46,11 @@ define(function(require) {
                 minDate: startDate,
                 maxDate: maxDate,
                 onShow: function(ct) {
+                    var start = self.$('.start').val();
```
eliottwiener-thirdchannel: Can't use `startDate` here?

josethirdchannel: In this context `startDate` is the first non-greyed out date on the calendar rather than the start of the date range in the UI. The naming here is not the best

PR #305, opened 2017-10-16
WF-250 - fixed issue where blackout scheme would occasionally fail

spullen:
https://thirdchannel.atlassian.net/browse/WF-250

* Fixed issue where, delete would result in an extra save, and a save would result in a duplicate save. This combination would occasionally cause the blackout scheme delete to fail.
* Did some cleanup of the view code

Also see PR for job service https://github.com/thirdchannel/jobs/pull/71


templates/handlebars/oddjob/blackout_schemes/blackout_scheme.hbs
```diff
@@ -1,14 +1,8 @@
...
-<div class="actions" id="blackout-scheme-save">
-	<button class="btn primary" id="blackout-scheme-delete">
-		<i class="ic fa ic_remove"></i>
-		<span>Delete</span>
-	</button>
-	<button class="btn primary" id="blackout-scheme-save">
-		<i class="ic fa ic_check"></i>
-		<span>Save</span>
-	</button>
+<div class="actions">
```
spullen: This was the culprit. The id `blackout-scheme-save` is being used twice and there's an event handler attached to it. So, when save was clicked, it would trigger two save actions. And, when delete was clicked, it would trigger a call to the save action as well.

PR #306, opened 2017-10-17
Wf 150

josethirdchannel:
https://thirdchannel.atlassian.net/browse/WF-150


templates/handlebars/oddjob/jobs/_job_fields.hbs
```diff
@@ -13,7 +13,12 @@
             {{/roles}}
         </select>    
     </div>
-    
+
+    <div class="form-group">
```
spullen: Wonder if you can just wrap the input inside the label so you don't need to add those additional classes.

Something like:
```
<div class="form-group">
  <label class="label">
    Allow Repeat Visits Within (days)
    <input class="input job-repeat-buffer" type="number" step="1" min="0" value="{{jobRepeatBuffer}" name="jobRepeatBuffer" placeholder="0" />
  </label>
</div>
```

josethirdchannel: Tried doing without the extra CSS classes but there doesn't seem to be a way around needing the the first part of the label to be displayed as a block and the second part displayed inline except by using </br>.

PR #309, opened 2017-10-27
WF-286 - adding the ability to tag extra images on survey

spullen:
https://thirdchannel.atlassian.net/browse/WF-286

Also see PR for ThirdChannel


templates/handlebars/shared/s3uploader/checkin_image.hbs
```diff
@@ -15,6 +15,16 @@
         <label>Description:</label>
         <input type="text" class="image_label" data-attribute="label" name="{{image_type}}[][label]" value="{{label}}">
     </div>
+    {{#if mediaTaggingEnabled}}
+        <div class="media-tags">
+            <select name="{{image_type}}[][media_tag_ids][]" class="image_media_tags" data-attribute="media_tag_ids" multiple="multiple">
+                <option value=""></option>
+                {{#each mediaTags}}
+                    <option value="{{this.[1]}}">{{this.[0]}}</option>
```
josethirdchannel: These seem like key value pairs. Would it be difficult to change these so that they are object properties or something that gives more context than array indices?

spullen: Initially had it like that... However, there's one annoying thing with this, we have this code and the code on thirdchannel doing ~almost the same thing, but slightly different. This is more of a compatibility thing then anything for the rails helper `options_for_select`.

PR #310, opened 2017-10-31
Samsung media tagging wf 285

josethirdchannel:
https://thirdchannel.atlassian.net/browse/WF-285

adds a tag management page to the survey builder for adding tags and marking them active


js/apps/singleNickel/views/tags/list_item.js
```diff
@@ -0,0 +1,52 @@
...
+
+    return Backbone.View.extend({
+        tagName: 'tr',
+        template: HandlebarsTemplates['singleNickel/tags/list_item'],
+        events: {
+            'change .toggleActive': 'toggleActive',
+            'click .addTag': 'addTag'
+        },
+        initialize: function() {
+            _.bindAll(this, 'toggleActive');
```
spullen: `_.bindAll` takes a list as the second argument, so you could combine these two lines into `_.bindAll(this, 'toggleActive', 'render');`

js/apps/singleNickel/views/tags/list_item.js
```diff
@@ -0,0 +1,52 @@
...
+            _.bindAll(this, 'render');
+            this.model.bind('change', this.render);
```
spullen: For views it's better to use `listenTo` because it will clean itself up when the view is removed.

```
this.listenTo(this.model, 'change', this.render);
```

js/apps/singleNickel/views/tags/list_item.js
```diff
@@ -0,0 +1,52 @@
...
+            var attributes = _.extend({tag: this.model}, this.model.toJSON());
+            this.$el.html(this.template(attributes));
+            this.$el.attr("data-tag", this.model.get("id"));
+            return this;
+        },
+        addTag: function (e) {
+            e.preventDefault();
+            e.stopPropagation();
+            var self = this;
+            this.model.set('name', this.$el.find('.tagName').val());
```
spullen: Shorthand for `this.$el.find('.some-selector')` is just `this.$('.some-selector')`

js/apps/singleNickel/views/tags/list_item.js
```diff
@@ -0,0 +1,52 @@
...
+            var confirmation = confirm('Are you sure you want to save this tag?');
+            if(confirmation) {
+                this.model.save().success(function(response) {
+                    self.remove();
```
spullen: Why remove? It'd make more sense if this just became a list view once editing is done.

js/apps/singleNickel/views/tags/list_item.js
```diff
@@ -0,0 +1,52 @@
...
+                    self.model.set('id', response.uuid);
+                    self.model.set('active', response.active);
```
spullen: Would these fields not automatically get populated from the response? Or is it a no-content response?

js/apps/singleNickel/views/tags/list_item.js
```diff
@@ -0,0 +1,52 @@
...
+        addTag: function (e) {
+            e.preventDefault();
+            e.stopPropagation();
+            var self = this;
+            this.model.set('name', this.$el.find('.tagName').val());
+            var confirmation = confirm('Are you sure you want to save this tag?');
+            if(confirmation) {
+                this.model.save().success(function(response) {
+                    self.remove();
+                    self.model.set('id', response.uuid);
```
spullen: You can write a parser on the model to handle this mapping.

js/apps/singleNickel/views/tags/list_item.js
spullen: Also wouldn't use self, you can just bind this to the function.

```javascript
this.model.save().success(function(response) {
  // ...  
  this.model.set('id', response.uuid);
  // ...
}.bind(this))
```

js/apps/singleNickel/views/tags/list_item.js
```diff
@@ -0,0 +1,52 @@
...
+                    self.collection.add(self.model);
+                }).fail(function() {
+                    context.trigger('error');
+                });
+            }
+        },
+        toggleActive: function(e) {
+            e.preventDefault();
+            e.stopPropagation();
+            this.model.attributes.options = {}; // causes a circular json error otherwise
```
spullen: How so?

js/apps/singleNickel/views/tags/list_item.js
```diff
@@ -0,0 +1,52 @@
...
+            this.model.set('active', !this.model.get('active'));
+            this.model.save().fail(function() {
```
spullen: You can use the `{patch: true}` option only to send the changed attributes, that might solve your other issue. `this.model.save({patch: true}).success(function() { ... }).fail(function() { ... })`

js/apps/singleNickel/views/tags/list.js
```diff
@@ -0,0 +1,65 @@
...
+
+    return Backbone.View.extend({
+        template: HandlebarsTemplates['singleNickel/tags/list'],
+        events: {
+            'click .addTag': 'addTag'
+        },
+        initialize: function() {
+            _.bindAll(this, 'renderTagItem');
+            this.listenTo(this.collection, 'add', this.renderTags);
+            this.model = new Model({customer: this.collection.options.customer});
```
spullen: What is this being used for, the require for it doesn't make sense.

js/apps/singleNickel/views/tags/list.js
```diff
@@ -0,0 +1,65 @@
...
+        },
+        addTag: function () {
+            var itemView = new ItemView({collection: this.collection, model: new Tag({customer: this.customer, active: false})});
+            this.$('#tag-list-items').append(itemView.render().el);
+        },
+        renderTags: function() {
+            if(this.collection.isEmpty()) {
+                this.$('#tag-list-items').html('<tr><td colspan="4">No Tags</td></tr>');
+            } else {
+                var tagIds = this.$('#tag-list-items').find('input').map(function (idx, el) {
```
spullen: What is all of this trying to do, seems unnecessary. Why can't you just loop through the collection and render each tag?

templates/handlebars/singleNickel/tags/list_item.hbs
```diff
@@ -0,0 +1,14 @@
+{{#if id}}
+    <td><a href="/tags/{{id}}">{{name}}</a></td>
```
spullen: Do we need a link here?

templates/handlebars/singleNickel/tags/list_item.hbs
```diff
@@ -0,0 +1,14 @@
+{{#if id}}
```
spullen: One thing you might want to try to do is make this an "editable" view. Meaning you can edit everything when it's an the editable state. Check out http://todomvc.com/examples/backbone/ it might give you an idea as to how to do that. I also think, once you get that it should clean up some of that funky code above and make things work a bit cleaner.

js/apps/singleNickel/views/tags/list.js
josethirdchannel: This was an effort to avoid re-rendering the subviews for each tag. It should now only render this list view only once and use an event to add append a new tag input

js/apps/singleNickel/views/tags/list_item.js
josethirdchannel: that fixed it!

templates/handlebars/singleNickel/tags/list_item.hbs
josethirdchannel: I'm guessing it would do something like `HandlebarsTemplates[this.templateName]` where the template name depends on the the ID field being set? And we have a list_item.hbs and list_item_edit.hbs? The code has become a little more manageable after the cleanup but I could add that to improve readability

spullen: One way to do this would be (this is just an example):

View:
```javascript
Backbone.View.extend({
  template: HandlebarTemplate['path/to/template'],
  events: {
     'click .edit': 'handleEdit',
     'click .cancelEdit': handleCancelEdit',
     'click .save': 'handleSave'
  },
  initialize: function() {
    this.listenTo(this.model, 'change:editing', this.render);
  },
  render: function() {
    this.$el.html(this.template(this.model.toJSON()));
    return this;
  },
  handleEdit: function(e) {
    e.preventDefault();
    this.model.set('editing', true);
  },
  handleEditCancel: function(e) {
    e.preventDefault();
    // revert any values that might have been updated (depending on the rest of the view/model works (for instance save -> failure -> cancel should retain the original values))
    this.model.set('editing', false);
  },
  handleSave: function(e) {
    e.preventDefault();
    // do save stuffz (apply any values from form to model, call save)
    // after save (on success), set editing to false
    // after save (on failure), display error if any
  }
});
```
Template:
```javascript
{{#if editing}}
<input type="text" value="{{name}}" />
<button class="save">Save</button>
<button class="cancelEdit">Cancel</button>
{{else}}
<span>{{name}}</span>
<button class="edit">Edit</button>
{{/if}}
```

Or you could split out the template into two parts and just determine what template to render.

js/apps/singleNickel/collections/tags.js
```diff
@@ -0,0 +1,11 @@
+define(function(require){
+    var BaseCollection = require('singleNickel/collections/base'),
+        Tag = require('singleNickel/models/tag');
+
+    return BaseCollection.extend({
```
spullen: I would just use Backbone.Collection.extend. The BaseCollection/Model are meant for the actual survey building parts (nesting, cloning, etc...), tag management doesn't have a need for that.

js/apps/singleNickel/models/tag.js
```diff
@@ -0,0 +1,23 @@
+define(function(require){
+    var _ = require('underscore'),
+        context = require('context'),
+        BaseModel = require('singleNickel/models/base');
+
+    return BaseModel.extend({
```
spullen: Same. BaseModel is meant for the survey building piece, it's not named properly. Just use Backbone.Model.

spullen:
Couple minor changes, but otherwise looks good!

js/apps/singleNickel/routers/main.js
```diff
@@ -79,6 +83,15 @@ define(function(require){
                 context.trigger('error');
             });
         },
+        listTags: function (customerId) {
+            var tags = new TagCollection({customerId: customerId});
```
spullen: I think you still need to specify an empty list of models (otherwise this will just create a model with `{customerId: customerId}`, so `var tags = new TagCollection([], {customerId: customerId})`. You also need to update the `initialize` method of the TagCollection to accept two parameters:
```javascript
define(function(require){
    var Backbone = require('backbone'),
        Tag = require('singleNickel/models/tag');

    return Backbone.Collection.extend({
        model: Tag,
        initialize: function (models, options) {
            this.customerId = options.customerId;
        },
        url: function() {
            return '/api/customers/'+this.customerId+'/tags';
        }
    });
});
```

spullen: Here's an example to illustrate the problem http://jsbin.com/pejohujuzo/1/edit?js,console

spullen: It probably "works", if you look at my example the wrong way does still "work".
According to the documentation the constructor is an array of models followed by extra options, see http://backbonejs.org/docs/backbone.html#section-101

PR #312, opened 2017-11-01
WF-287 - adding support for Survey::QuestionImage

spullen:
https://thirdchannel.atlassian.net/browse/WF-287

* Some minor cleanup
* Adding survey answer update when an image has been submitted for a question based image

The whole survey flow isn't ideal and really needs to be redone, but until then this just tacks on question awareness to the image/file management part. The flow is, image is upload, once uploaded the image object is created, if all that works then it is made part of the answer to the question. First two steps are just the current process for any image.

Also see PR for ThirdChannel

Ended up separating the auto-tagging feature into its own ticket to move this along quicker.


js/apps/shared/views/s3uploader/checkin_file.js
```diff
@@ -87,41 +103,56 @@ define(function(require) {
...
                 }
 
-                self.$viewer.append(view.$el);
-            }).fail(function() {
-                self.fileUploadError();
-            });
+                // If this is an Survey::QuestionImage add hidden field to track id value
+                if(this.imageType.includes("question")) {
+                    var answer = document.createElement("option");
+                    answer.innerText = image.id;
```
josethirdchannel: is this necessary for a hidden field?

spullen: Probably not, but can't hurt either.

PR #313, opened 2017-11-03
WF-314 - updating when the media tags are displayed

spullen:
https://thirdchannel.atlassian.net/browse/WF-314

Also see PR for thirdchannel


js/apps/shared/views/s3uploader/checkin_file.js
```diff
@@ -27,9 +27,16 @@ define(function(require) {
             var self = this;
 
             this.imageType = self.$el.data('image-type');
+            this.displayMediaTags = this.$el.data('display-media-tags');
             this.$form = $('.s3_uploader');
             this.$viewer = this.$('.viewer');
 
+            console.log('displayMediaTags', this.displayMediaTags);
```
josethirdchannel: this should be safe to remove

PR #321, opened 2017-11-17
Tag editing

josethirdchannel:
allows clicking on a tag name to edit the text field and adds a save button


js/apps/singleNickel/views/tags/list_item.js
```diff
@@ -34,6 +36,34 @@ define(function(require) {
...
+                .attr('type', 'text')
+                .addClass('tagName');
+
+            var saveButton = $('<a></a>')
+                .attr('href', '#')
+                .addClass('btn light saveTag')
+                .html('save');
+            this.$('.name').empty();
+            this.$('.name').append(inputEl).append(saveButton);
+            this.$el.off('click', '.name');
```
spullen: is it necessary to turn this off and on

josethirdchannel: the listener is on the outer div that contains the input field or tag name display. Without disabling this it'll clear what the user has typed on click by deleting the current input field and appending a new blank one

spullen: This is where having different show/edit templates and rendering them (or portions of the view) depending on the state.

This is fine though.

PR #324, opened 2017-12-08
Fix formatting when re-opening video window

dfox:


js/libs/viewer.js
```diff
@@ -1088,7 +1088,7 @@
       }
 
       $item.addClass(CLASS_ACTIVE);
-      this.$canvas.toggleClass('video', video);
```
spullen: Seems like this should work? toggleClass with true would add the class 'video', false would remove it if present.

